PR #4, opened 2015-04-03
Deploy an engage node. Includes:

djcp:
* Update host-dependent configurations during the configure lifecycle
 event
* Set the admin hostname during the configure event
* Build first, then configure to avoid configuration race conditions


recipes/deploy-engage.rb
```diff
@@ -0,0 +1,101 @@
+# Cookbook Name:: mh-opsworks-recipes
```
djcp: Lots of duplication amongst these deploy recipes, once they're relatively stable I think a refactor is in order.

PR #27, opened 2015-11-02
Use an asset server when available for external file downloads.

djcp:
This helps configure matterhorn to use an asset server to seed the CDN origin.
The asset server is a tuned nginx machine that serves large static files
efficiently. Using the asset server to populate the CDN takes significant IO
load off the engage server.

Asset server selection logic:
* If there's a cloudfront_url, use that and emit it from engage.
* If there's a configured asset server, use that. This is now the
 default.
* If there's neither, jpgs and videos are delivered directly from the
 engage node. This is OK for dev clusters.

### DIFF
--- a/templates/default/engage-nginx-proxy-conf.erb
+++ b/templates/default/engage-nginx-proxy-conf.erb
@@ -1,6 +1,7 @@
 server {
   listen 80 default_server;
   listen [::]:80 default_server ipv6only=on;
+  directio 1M;
 
   root /usr/share/nginx/html;
   index index.html index.htm;
@@ -30,6 +31,7 @@ server {
 server {
   listen 443 ssl;
   listen [::]:443 ipv6only=on ssl;
+  directio 1M;
 
   root /usr/share/nginx/html;
   index index.html index.htm;
--- a/libraries/default.rb
+++ b/libraries/default.rb
@@ -56,6 +56,17 @@ module MhOpsworksRecipes
       admin_hostname
     end
 
+    def using_asset_server?
+      node[:opsworks][:layers][:asset_server]
+    end
+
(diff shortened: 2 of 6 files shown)


templates/default/engage-nginx-proxy-conf.erb
```diff
@@ -1,6 +1,7 @@
 server {
   listen 80 default_server;
   listen [::]:80 default_server ipv6only=on;
+  directio 1M;
```
lbjay: So if I understand correctly this bypasses OS caching for anything > 1M. And we want this because caching is the responsibility of cloudfront? Is 1M low enough to cover everything?

djcp: This optimization exists to avoid a lot of buffer cache churn.  Unless you have a machine with an ungodly amount of RAM or a situation where it serves the same large file over and over again, the buffer cache is going to have a very low hit rate for large video files.

I set it to 1M so that everything under that - the JPG thumbnails - would be stored in the buffer cache but none of the video files would. 

In reality, if we've tuned cloudformation correctly we shouldn't need to care about this at all.  If the setting is wrong, it's just going to lead to more files being served directly off the filesystem.

I went back-and-forth on this but basically every nginx tutorial about tuning large files suggested managing the directio setting. It seems pretty low risk especially since this is just going to be a CDN seed server.

djcp: And 1M was based on a `find` command I ran that found no JPG thumbnails were larger than that on the prod FS.

lbjay: cool beans. thanks for filling in my blanks.

lbjay:
:+1:

PR #31, opened 2015-12-02
Use new mo-scaler scaling manager. Includes:

lbjay:
* updated repo/version
    * new cron entries for normal, offpeak and weekend operation

### DIFF
--- a/recipes/install-ec2-scaling-manager.rb
+++ b/recipes/install-ec2-scaling-manager.rb
@@ -3,7 +3,18 @@
 
 ::Chef::Recipe.send(:include, MhOpsworksRecipes::RecipeHelpers)
 
-manager_release = node.fetch(:ec2_management_release, 'v0.2.2')
+moscaler_attributes = {
+  ec2_management_release: 'v0.2.2',
+  offpeak_instances: 2,
+  peak_instances: 10,
+  weekend_instances: 1
+}.merge(node.fetch(:moscaler, {}))
+
+manager_release = moscaler_attributes[:ec2_management_release]
+offpeak_instances = moscaler_attributes[:offpeak_instances]
+peak_instances = moscaler_attributes[:peak_instances]
+weekend_instances = moscaler_attributes[:weekend_instances]
+
 rest_auth_info = get_rest_auth_info
 stack_name = node[:opsworks][:stack][:name]
 region = "us-east-1"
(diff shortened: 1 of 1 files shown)


recipes/install-ec2-scaling-manager.rb
```diff
@@ -25,33 +26,55 @@
...
-cron_d 'ec2_manager' do
+# weekdays, offpeak, every five minutes from midnight - 7am + 11pm - midnight
+cron_d 'moscaler_offpeak' do
   user 'ec2_manager'
-  minute '*/2'
-  command %Q(cd /home/ec2_manager/ec2-management && /usr/bin/run-one ./ec2_manager.py --opsworks "#{stack_name}" autoscale 2>&1 | logger -t info)
+  hour '0-7,23'
+  minute '*/5'
+  weekday '1-5'
+  command %Q(cd /home/ec2_manager/mo-scaler && /usr/bin/run-one ./manager.py scale to 2 2>&1 | logger -t info)
```
djcp: "scale to <integer>" - neat!

recipes/install-ec2-scaling-manager.rb
```diff
@@ -25,33 +26,55 @@
...
+  path '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin'
+end
+
+# weekdays, normal production window, every five minutes from 8am - 11pm
+cron_d 'moscaler_normal' do
+  user 'ec2_manager'
+  hour '8-22'
+  minute '*/5'
+  weekday '1-5'
+  command %Q(cd /home/ec2_manager/mo-scaler && /usr/bin/run-one ./manager.py scale to 5 2>&1 | logger -t info)
```
djcp: We only run 5 workers during max production?

djcp:
Looks great! One question.

recipes/install-ec2-scaling-manager.rb
lbjay: Oh, hah. Confirming what these numbers should be.

djcp:
Actually - another question. I'm thinking that we'll deploy most dev clusters with only 2 workers by default to save $$. As it is now, this'd try to scale up to an invalid number of instances during the weekdays. Will that be OK?  More generally, I'm asking if a mismatch between the scaler and the number of available instances will be an issue or not.

lbjay:
If a dev cluster had only 2 workers and was asked to scale to 10 the operation would abort with a log entry like "not enough workers". Related: the scale up/down operations are all-or-nothing, ie, if a cluster had 2 online workers out of a total of 5 workers and was asked to scale to 10, it would not scale up to 5.

PR #33, opened 2015-12-18
newrelic integration. Includes:

djcp:
* Install the newrelic configuration
* Inflect the names from the node name and a user supplied
 "environment_name"
* Inject newrelic into the matterhorn init script when it's enabled
* Configure newrelic logging

### DIFF
--- a/templates/default/newrelic.yml.erb
+++ b/templates/default/newrelic.yml.erb
@@ -1,0 +1,295 @@
+# This file configures the New Relic Agent.  New Relic monitors
+# Java applications with deep visibility and low overhead.  For more details and additional
+# configuration options visit https://docs.newrelic.com/docs/java/java-agent-configuration.
+#
+# This configuration file is custom generated for Harvard Division of Continuing Education_1
+#
+# This section is for settings common to all environments.
+# Do not add anything above this next line.
+common: &default_settings
+
+  # ============================== LICENSE KEY ===============================
+  # You must specify the license key associated with your New Relic
+  # account. For example, if your license key is 12345 use this:
+  # license_key: '12345'
+  # The key binds your Agent's data to your account in the New Relic service.
+  license_key: "<%= @newrelic_key %>"
+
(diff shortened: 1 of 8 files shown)


djcp:
Assuming your stack is named "stack name", the nodes end up logging stuff to newrelic under "stack name" and "stack_name_admin1" (etc.) so that we get aggregate and individual node stats.

templates/default/newrelic.yml.erb
```diff
@@ -0,0 +1,295 @@
...
+  class_transformer:
+    # This instrumentation reports the name of the user principal returned from 
+    # HttpServletRequest.getUserPrincipal() when servlets and filters are invoked.
+    com.newrelic.instrumentation.servlet-user:
+      enabled: false
+
+    com.newrelic.instrumentation.spring-aop-2:
+      enabled: false
+
+  # Phoebe: For pricing
```
gregorydlogan: Some clarity around what these keys are and what they do would be helpful.

templates/default/newrelic.yml.erb
```diff
@@ -0,0 +1,295 @@
...
+  # Limits the number of lines to capture for each stack trace. 
+  # Default is 30
+  max_stack_trace_lines: 30
+
+  # Provides the ability to configure the attributes sent to New Relic. These
+  # attributes can be found in transaction traces, traced errors, Insight's 
+  # transaction events, and Insight's page views.
+  attributes:
+  
+    # When true, attributes will be sent to New Relic. The default is true.
```
gregorydlogan: This file appears to have a variety of different indentation levels, is that deliberate?

djcp: Yup! It's a YAML file.  The indentation denotes key nesting.

templates/default/newrelic.yml.erb
djcp: it's all newrelic config specific, so it seems a little odd to document it here.

PR #35, opened 2016-01-08
Fix the matterhorn init script to restart until actually successful

djcp:
Background:

There's a race condition that causes the daemon to not restart correctly about
10 to 15% of the time. The easiest check is to see if the login appears after
a restart. This modifies the init process to look for
"j_username" in the source of the homepage, which means we're past the race
condition and the daemon is up.

If it can't succeed in 10 attempts, it aborts.

### DIFF
--- a/templates/default/matterhorn-init-script.erb
+++ b/templates/default/matterhorn-init-script.erb
@@ -30,6 +30,13 @@ lockfile=/var/lock/subsys/matterhorn
 [ -d "/var/lock/subsys" ] || lockfile="/var/lock/LCK.${prog}"
 pidfile="/var/run/${prog}.pid"
 
+# Used in the "restart until the daemon is actually working" feature
+# now included in start()
+test_string="j_username"
+max_start_attempts=10
+attempts=0
+
+
 killdelay=7
 
 # Load configuration files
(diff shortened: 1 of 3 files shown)


templates/default/matterhorn-init-script.erb
```diff
@@ -59,6 +66,20 @@ start() {
...
 
+  # Attempt to restart the daemon until it actually looks like it's
+  # started correctly
+  sleep 10
+  if [ "$attempts" -lt "$max_start_attempts" ]
+    then
+      if ! $(/usr/bin/curl -s -L http://localhost/ | grep -q "$test_string"); then
+        attempts=$[$attempts + 1]
+        restart
+      fi
```
spiraleddy: Dropping a log line about the restart would identify start up issues here. It would also identify if the 10 seconds was too soon to be checking.

djcp: This does get logged into the chef logs from opsworks, IIRC. Otherwise, yeah, I can throw something into syslog but I want to test it first.

spiraleddy:
This is good to automate failed restarts because of the IllegalStateException/Whiteboard error in Jetty 6. I'm actually Ok with it not dropping a log line for now, because the IllegalStateException should still show up in the Matterhorn logs to identify if this issue is still happening. So Ok by me

PR #36, opened 2016-01-11
Send X-Forwarded-For header so that user-tracking gets correct client ip

lbjay:
Verified correct client IP is received using jluker-ngnix-x-forwarded-for-header cluster.


djcp:
Looks good! I will merge for you, I am working on a new release process now that we've moved to tagged releases and I started on it after you opened this branch.

PR #41, opened 2016-01-15
MATT-1847 hot fix Otherpubs config tag

spiraleddy:
Hi Dan, can you review this hotfix for the current production deploy. The config tags changed for Otherpubs and this file puts the config in synch.


djcp:
Looks good to me! Can you tag and merge this into the relevant branches in this repo? After that, we'd:

1) Update the tag in the cluster config,
2) Update the cluster-wide chef recipes

and given that we've already manually made this change I don't think we need to redeploy.

spiraleddy:
Ok, I'm going to 
1. merge this to master push, 
2. create tag off master as v1.0.5, and push
3.  merge tag v1.0.5 to develop

PR #42, opened 2016-01-19
f/matt 1727 distribute to s3

rute-santos:
Added configuration for the new Matterhorn S3 distribution service (bucket name and region).

### DIFF
--- a/libraries/default.rb
+++ b/libraries/default.rb
@@ -471,6 +471,18 @@ module MhOpsworksRecipes
       end
     end
 
+    def install_aws_s3_distribution_service_config(current_deploy_root, region, s3_distribution_bucket_name)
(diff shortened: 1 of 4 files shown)


libraries/default.rb
```diff
@@ -471,6 +471,18 @@ def install_live_streaming_service_config(current_deploy_root,live_stream_name)
...
+        source 'edu.harvard.dce.distribution.aws.s3.AwsS3DistributionServiceImpl.properties.erb'
+        owner 'matterhorn'
+        group 'matterhorn'
+        variables({
+          region: region,
+          s3_distribution_bucket_name: s3_distribution_bucket_name 
+        })
+      end
+    end
+
```
djcp: Will all clusters (dev or not) require an s3 distribution bucket name from now on?  

Put another way - if this service isn't configured, will a cluster "just work?" If not, then I'll need to think about backwards compatibility and how to automate bucket creation during creation.

PR #43, opened 2016-01-21
Distribute directly from s3, not the engage node

djcp:


lbjay:
:+1:

templates/default/config.properties.erb
```diff
@@ -86,10 +86,9 @@ org.opencastproject.download.directory=${org.opencastproject.shared_storage.dir}
 # The base URL for media downloads.
 <% if @cloudfront_url %>
 org.opencastproject.download.url=https://<%= @cloudfront_url %>
-<% elsif @using_asset_server %>
-org.opencastproject.download.url=http://<%= @asset_server_hostname %>/static
 <% else %>
-org.opencastproject.download.url=<%= (@using_ssl_for_engage) ? 'https://' : 'http://' %><%= @engage_hostname %>/static
+# Distribute directly from the s3 bucket.
```
rute-santos: I believe the virtual-hosted-style url is preferred: https://<%= @s3_distribution_bucket_name %>.s3.amazonaws.com

djcp: nice catch! fixed.

djcp:
oops, this should've been issued against develop (though it's the same as master).  Merged!

PR #44, opened 2016-01-21
Usertracking session + uid log for reconciling anonymous shopper user actions

lbjay:
This updates the engage nginx config to set a unique "uid" cookie which on subsequent requests is logged to a new access log containing two values: the user's usertracking session id (JSESSIONID) and the tracking cookie. This log will later be used to associate "anonymous" user tracking events created during the shopping period with future events from those same users should they become registered.

Note: I debated about the uid cookie's expiration and decided to just go with "max" which means it basically never expires. We really only need it to persist past the point where it's possible for new students to register (i.e. a couple of weeks), but then I didn't want to rule out other, unanticipated ways it could be useful. Please express your strong feelings if you have them.

### DIFF
--- a/templates/default/engage-nginx-proxy-conf.erb
+++ b/templates/default/engage-nginx-proxy-conf.erb
@@ -1,3 +1,9 @@
+userid on;
+userid_name dce_uid;
+userid_expires max;
+
+log_format session_uid '$cookie_JSESSIONID $uid_got';
+
 server {
   listen 80 default_server;
   listen [::]:80 default_server ipv6only=on;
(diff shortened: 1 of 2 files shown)


templates/default/engage-nginx-proxy-conf.erb
```diff
@@ -26,8 +33,12 @@ server {
     alias <%= @shared_storage_root %>/downloads;
   }
 
+  location /usertracking {
+    proxy_pass http://127.0.0.1:<%= @matterhorn_backend_http_port %>/usertracking;
+    access_log /var/log/nginx/session_uid.log session_uid;
```
djcp: Does this interfere with normal nginx access logging at all?

lbjay: Turns out the answer was yes. Updated to repeat the base config's standard access_log setting.

PR #45, opened 2016-01-22
Rotate nginx logs daily, compressing immediately

djcp:


djcp:
The default nginx logrotate config only rotates weekly, keeping for a year. This means we have around a gig of hot logs on the engage node now, which could be a problem when we start to see traffic. This switches to rotating daily, compressing immediately and still keeps a year. Deploying this requires no interruption to any running service.

djcp:
merged into develop. Man, I need to fix how I'm doing PRs.

PR #48, opened 2016-02-01
Create an rsync-only "file_uploader" user.

djcp:
This allows you to get files directly into the matterhorn inbox via rsync.
This allows uploads to be scripted easily.

To use this feature:
* create an ssh keypair,
* add the public key to your cluster config's "custom_json" section,
* invoke rsync from your clients that're uploading files:

        rsync --progress -auz -e "ssh -i <path to the private ssh key for this
user>" ~/sample_video/file_to_upload.avi file_uploader@<your node's public
hostname>:/var/matterhorn/rsync_uploads/

Includes:

* Install the "file_uploader" user with an authorized_key defined by
 "file_uploader_public_ssh_key" in your `custom_json`
* Limit the file_uploader user to only be able to run rsync in the
 nfs-mounted "rsync_uploads" directory.
* Install a cron job to move files from the `rsync_uploads` directory into
 the default matterhorn inbox directory.
* Install an hourly cron job to remove rsync uploads older than 24 hours
 to catch stale uploads.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,16 @@
         # This is currently on the "Ganglia" layer in prod,
         # the value below is for dev clusters.
         ./bin/rake stack:commands:execute_recipes_on_layers layers="Admin" recipes="mh-opsworks-recipes::install-mysql-backups"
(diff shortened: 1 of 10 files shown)


djcp:
merged into develop

PR #49, opened 2016-02-03
Make the java debug setting tweakable, defaulting to "true"

djcp:


lbjay:
:+1:

spiraleddy:
Dan, but the production defaults to FALSE, right? To prevent remote connections in production.

PR #55, opened 2016-02-16
Copy and symlink nginx log dir if it's not already a symlink

djcp:


djcp:
This would be run on the engage node only, which gets super big nginx logs because of the usertracking stuff.

djcp:
Yeah, I forget and then remember it every couple months.

PR #56, opened 2016-02-18
Install awscli directly via pip. Includes:

djcp:
* Use our homegrown awscli installer instead of the upstream chef recipe
* Tie to a specific awscli release
* Test the awscli install and specified release


lbjay:
:+1:

PR #57, opened 2016-02-19
Configure a squid proxy for the storage IP.

djcp:
This recipe should be run on an instance with a public IP address, and acts as
a squid proxy for zadara s3 snapshots. Details here:

https://support.zadarastorage.com/entries/69891364-Setup-Backup-To-S3-B2S3-Through-a-Proxy-In-Your-AWS-VPC

Includes:

* Extract get_storage_hostname from nfs-client
* Configure squid3 to allow access to the various s3 endpoints from the
 resolved VPSA or storage IP address.

### DIFF
--- a/libraries/default.rb
+++ b/libraries/default.rb
@@ -173,6 +173,19 @@ module MhOpsworksRecipes
       )
     end
 
+    def get_storage_hostname
+      storage_info = get_storage_info
+
+      if storage_info[:type] == 'external'
+        storage_info[:nfs_server_host]
+      else
+        layer_shortname = storage_info[:layer_shortname]
+        (storage_hostname, storage_available) = node[:opsworks][:layers][layer_shortname.to_sym][:instances].first
+
+        storage_hostname
+      end
+    end
+
     def get_shared_storage_root
       storage_info = get_storage_info
       storage_info[:shared_storage_root] || storage_info[:export_root]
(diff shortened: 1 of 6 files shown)


libraries/default.rb
```diff
@@ -173,6 +173,19 @@ def get_storage_info
       )
     end
 
+    def get_storage_hostname
+      storage_info = get_storage_info
+
+      if storage_info[:type] == 'external'
```
lbjay: Assuming it's not a problem that the `:type` symbol is not defined in the default return from get_storage_info.

recipes/nfs-client.rb
```diff
@@ -5,13 +5,7 @@
...
 include_recipe 'mh-opsworks-recipes::create-metrics-dependencies'
 
-storage_info = node.fetch(
-  :storage, {
-    export_root: '/var/tmp',
-    network: '10.0.0.0/8',
-    layer_shortname: 'storage'
-  }
-)
+storage_info = get_storage_info
```
lbjay: The diffs here are a bit confusing as it seems `get_storage_info` was previously already defined but not being used here for some reason.

libraries/default.rb
djcp: nope, because it'd be `nil`.

recipes/nfs-client.rb
djcp: It was defined before, I just wasn't using it here. I am just deduplicating the logic.

lbjay:
:+1:

djcp:
So I need to wait for an available zadara mount before I can actually test this squid proxy - the goal is to be able to set up the object store dumps off the zadara array.

djcp:
Tested! Works fine

PR #58, opened 2016-02-22
Added 401 to ignore

phoebem:
This is missing from the newrelic template - it removes the authentication object errors

[ticket: X]

### DIFF
--- a/templates/default/newrelic.yml.erb
+++ b/templates/default/newrelic.yml.erb
@@ -197,7 +197,7 @@ common: &default_settings
     # by providing a comma separated list of status codes.
     # The default is to exclude 404s. If you want to override
     # this, you must provide any new value as an empty list is ignored.
-    ignore_status_codes: 404
+    ignore_status_codes: 404,401
 
   # Transaction Events are used for Histograms and Percentiles. Unaggregated data is collected
   # for each web transaction and sent to the server on harvest. 


djcp:
Is this treating the symptom more than the illness? I guess my concern is that we might want to see legitimate 401s in newrelic as a way to find, say, persistent authentication attacks or whatever. 

Would it be difficult to fix this at the matterhorn level instead of ignoring the status codes all together?

phoebem:
We had investigated this before and this is the normal way the authentication works.

Phoebe.

phoebem:
Challenge-Response Handshake 
The client does not send the Authorization header when sending its request to the server (it does not know that the server requires HTTP Basic Authentication). The server responds with an HTTP 401 response code, instructing the client to authenticate to the server by sending the Authorization header. The client then sends a second request, this time including the Authorization header and the relevant username and password.

djcp:
Couldn't we just make the client use the correct authentication method on
the first request? Or does the first request include the token the client
needs to use in its response?

On Mon, Feb 22, 2016 at 2:54 PM, phoebe miller <notifications@github.com>
wrote:

> Challenge-Response Handshake
> The client does not send the Authorization header when sending its request
> to the server (it does not know that the server requires HTTP Basic
> Authentication). The server responds with an HTTP 401 response code,
> instructing the client to authenticate to the server by sending the
> Authorization header. The client then sends a second request, this time
> including the Authorization header and the relevant username and password.
>
>
> > On Feb 22, 2016, at 2:50 PM, Phoebe Miller <pmiller@dce.harvard.edu>
> wrote:
> >
> > We had investigated this before and this is the normal way the
> authentication works.
> >
> > Phoebe.
> >
> >> On Feb 22, 2016, at 2:26 PM, Daniel Collis-Puro <
> notifications@github.com <mailto:notifications@github.com>> wrote:
> >>
> >> Is this treating the symptom more than the illness? I guess my concern
> is that we might want to see legitimate 401s in newrelic as a way to find,
> say, persistent authentication attacks or whatever.
> >>
> >> Would it be difficult to fix this at the matterhorn level instead of
> ignoring the status codes all together?
> >>
> >> —
> >> Reply to this email directly or view it on GitHub <
> https://github.com/harvard-dce/mh-opsworks-recipes/pull/58#issuecomment-187331720
> >.
>
> >>
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/harvard-dce/mh-opsworks-recipes/pull/58#issuecomment-187352765>
> .
>



-- 
Dan Collis-Puro
Principal Engineer, Division of Continuing Education
http://www.dce.harvard.edu  :: @djcp :: github.com/djcp

djcp:
Background from a convo between tobias and greg, via phoebe:

Hi Greg,

good to see that your moving to application profiling, too!

In the New Relic configuration there is a setting that allows suppression of specific HTTP error codes. We are by default filtering out 401, which is an "error" code you get as part of how Digest Authentication is implemented. So we did the same thing you did (but did not see anything "bad" :-).

New Relic is designed to report HTTP status codes beyond the 3xx range, and since this 4xx is part of the protocol, I would suggest it's ok to filter it out.

Tobias

On 09 Nov 2015, at 18:43, Greg Logan <gregorydlogan@gmail.com<mailto:gregorydlogan@gmail.com>> wrote:

Hi folks,

At DCE we're just starting to roll out NewRelic on our servers, but we're seeing lots of errors across the entire system.  The major culprit of these errors appears to be the normal digest auth mechanism.  At the beginning of nearly every HTTP transaction between nodes there is a 401 returned as part of normal operations.  This gets picked up as an error by NewRelic at the rate of roughly 20 errors/minute/node - not good :)

We have since figured out how to ignore these errors, but that seems like bad form in my eyes.  I know there is at least one other institution in the community using NewRelic, so I'm hoping to draw on some experience here:  How did you address this issue?  Were there other issues that cropped up that you can think of?

Thanks,
G

djcp:
I'm not happy with the fix and would prefer the behavior was fixed, but whatever.

LGTM. Please rebase against develop (if needed), add an entry into the CHANGELOG.md and merge.

djcp:
In the future, please add an entry to CHANGELOG.md.

phoebem:
This is the correct behavior.
We use digest authentication.

I can type it up, but they do it better, see:

https://en.wikipedia.org/wiki/Digest_access_authentication

spiraleddy:
There is no way to distinguish a 401 response for a digest client and a non-digest client. The server sends the nonce and "WWW-Authenticate Digest" in the 401 header in both cases. It might be possible to detect unauthorized clients by counting the 401s & the 200s sent from the server. If outbound 401s exceed the 200s, then clients without digest auth are probably attempting access.

djcp:
I guess my complaint is mostly with digest auth, which does one too many
HTTP requests per transaction. I'm used to token auth where you have a
shared secret between your clients and servers that you communicate over
HTTPS, and which doesn't require an extra HTTP request.

It just seems silly to turn 20 requests into 40, half of which are failures
by design - with all the associated log entries, load (minor, but still
there) and complexity. But if that's how matterhorn handles auth and
there's no easy fix to use token auth (or no auth from the private
network), so be it.

--DJCP

On Tue, Feb 23, 2016 at 8:38 AM, Karen Dolan <notifications@github.com>
wrote:

> There is no way to distinguish a 401 response for a digest client and a
> non-digest client. The server sends the nonce and "WWW-Authenticate Digest"
> in the 401 header in both cases. It might be possible to detect
> unauthorized clients by counting the 401s & the 200s sent from the server.
> If outbound 401s exceed the 200s, then clients without digest auth are
> probably attempting access.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/harvard-dce/mh-opsworks-recipes/pull/58#issuecomment-187703247>
> .
>



-- 
Dan Collis-Puro
Principal Engineer, Division of Continuing Education
http://www.dce.harvard.edu  :: @djcp :: github.com/djcp

spiraleddy:
It's possible for Matterhorn use SSL directly, although, in the past, Tobias and others recommend leaving Matterhorn on HTTP (with the digest) and wrapping with apache/ngix SSL. The following posts contain interesting arguments on each approach and links to wiki pages for each approach (for older versions of Opencast).

In the first, Alexander describes problems from trying a mixture of both,  http://opencast.3480289.n2.nabble.com/Matterhorn-users-Matterhorn-behind-a-Apache-proxy-with-SSL-td7583376.html. Tobias recommended keeping Matterhorn on HTTP (with digest auth) and using apache/ngix SSL. 

In the second post, Ruth wants to use SSL directly and Stephen's counter concern https://groups.google.com/a/opencast.org/forum/#!topic/matterhorn-users/piYt2PG2hZY.

PR #59, opened 2016-02-22
Switch to daily full mysqldump backups

djcp:
This is because RDS has everything we need in terms of point-in-time backups.
These daily dumps are more for historical record.


lbjay:
:+1:

PR #60, opened 2016-02-22
Create a default .bashrc that includes a colorized prompt with the cluster name

djcp:


djcp:
screenshot-2016-02-22t16 21 16
screenshot-2016-02-22t16 20 01

djcp:
The bashrc file is huge because it's the default ubuntu 14.04 bashrc plus our custom prompt and a fix for terminal color init (both noted with inline comments).

This doesn't effect root's prompt, which apparently doesn't use `/etc/skel/.bashrc`  I should probably fix that one, too.

lbjay:
:+1:

PR #64, opened 2016-03-02
Make it easier to deploy matterhorn into a subdirectory.

djcp:
Create a recipe to allow you to remove a deployment root directory across all
nodes - this is dangerous and should not be run routinely.


lbjay:
LGTM :+1:

PR #67, opened 2016-03-08
Create and apply cluster seed files

djcp:


files/default/create-cluster-seed-file.sh
```diff
@@ -0,0 +1,119 @@
...
+  echo
+  echo "Usage: create-cluster-seed-file.sh -x -p <full path to shared matterhorn files>"
+  echo
+  echo "This creates a tarball at <shared matterhorn file path>/<your seed cluster name>.tgz"
+  echo "of the mysql data and matterhorn-generated files necessary to recreate"
+  echo "this cluster in another environment. This is not meant as a backup, it is tooling"
+  echo "to make automated testing easier. After the seed file is created, it is uploaded"
+  echo "to the bucket defined as --upload_s3_bucket in the options."
+  echo
+  echo " -x	Actually create the backup"
```
lbjay: Is "backup" the right term here?

djcp: Nope. Will fix!

files/default/remove-solr-indexes.sh
```diff
@@ -0,0 +1,38 @@
...
+    *)
+      break
+  esac
+
+  shift
+done
+
+if ! $do_it; then
+  echo
+  echo "Usage: remove-engage-indexes.sh -x -p <full path to indexes>"
```
djcp: Fix

recipes/deploy-admin.rb
```diff
@@ -116,7 +113,6 @@
         s3_distribution_bucket_name: s3_distribution_bucket_name,
         capture_agent_monitor_url: capture_agent_monitor_url,
         live_streaming_url: live_streaming_url,
-        using_ssl_for_engage: using_ssl_for_engage,
```
lbjay: is this a residual change from earlier refactoring? I can't seem to trace the relevance.

djcp: Yes, I can see how this would be confusing. It is a fix to the engage integration that I missed when we switched to s3 distribution - I am passing the `using_ssl_for_engage` attribute to the template but not using it any more.  I caught it in this branch because of the work I was doing to restore a cluster's files across s3 buckets.

files/default/create-cluster-seed-file.sh
```diff
@@ -0,0 +1,119 @@
...
+    *)
+      break
+  esac
+
+  shift
+done
+
+if ! $do_it; then
+  echo
+  echo "Usage: create-cluster-seed-file.sh -x -p <full path to shared matterhorn files>"
```
djcp: Update usage info

files/default/load-seed-data.sh
```diff
@@ -0,0 +1,79 @@
...
+      ;;
+    *)
+      break
+  esac
+
+  shift
+done
+
+if ! $do_it; then
+  echo
```
djcp: Update usage info

files/default/modify-database-after-loading-seed-data.sh
```diff
@@ -0,0 +1,116 @@
...
+      break
+  esac
+
+  shift
+done
+
+source "$shared_files_path/seed_cluster_hostnames.txt"
+
+if ! $do_it; then
+  echo
```
djcp: ditto - update usage info

files/default/load-seed-data.sh
```diff
@@ -0,0 +1,79 @@
...
+
+rm -Rf *
+aws s3 cp s3://"$bucket_name"/"$seed_file" .
+tar xvfz "$seed_file"
+rm "$seed_file"
+
+aws s3 sync ./s3_contents/ s3://"$s3_distribution_bucket" --delete
+
+/usr/bin/mysql -e 'drop database matterhorn; create database matterhorn;'
+/usr/bin/mysql matterhorn < "$shared_files_path/mysql_seed_backup/matterhorn.mysql"
```
lbjay: Wondering if it would be worth it to "parallelize" the s3 sync and the mysql load procs. Would it be as simple as `&` backgrounding them and then a `wait` at the end?

djcp: We probably could do that - the database restore probably takes almost no time at all, though. It only takes a lot of time in production because of all the user stats.

lbjay: OK, maybe not worth it. I only had the chef log to go on which showed the "execute[load seed database and create seed files]" block taking 2.5m of the total 4m when applying the seed.

lbjay:
:+1: lgtm

PR #68, opened 2016-03-22
loggly TLS cert file updated

lbjay:
as per https://www.loggly.com/docs/upgrade-tls-certificate/.

This needs to be deployed prior to March 30, 2016 at 10:00 AM PST.


djcp:
:+1:

PR #70, opened 2016-03-29
Improvements to allow for local opsworks deployments via vagrant.

djcp:


libraries/default.rb
```diff
@@ -39,6 +39,15 @@ def get_seed_file
       node.fetch(:cluster_seed_file, 'cluster_seed.tgz')
     end
 
+    def on_aws?
+      if node['vagrant_environment'] == true
+        Chef::Log.info "deploying to a vagrant cluster"
```
lbjay: This logs every time `on_aws?` gets called?

djcp: oops, I issued this against the wrong branch. I've opened #71 against the correct branch.

libraries/default.rb
```diff
@@ -469,11 +478,13 @@ def download_episode_defaults_json_file(current_deploy_root)
         recursive true
       end
 
-      execute 'download the EpisodeDefaults.json file into the correct location' do
-        command %Q|cd #{episode_default_storage_dir} && aws s3 cp s3://#{private_assets_bucket_name}/EpisodeDefaults.json EpisodeDefaults.json|
-        retries 10
-        retry_delay 5
-        timeout 300
+      if node['vagrant_environment'] != true
```
lbjay: `on_aws?` ?

PR #71, opened 2016-03-29
Changes to support a local vagrant opsworks cluster.

djcp:


libraries/default.rb
```diff
@@ -39,6 +39,15 @@ def get_seed_file
       node.fetch(:cluster_seed_file, 'cluster_seed.tgz')
     end
 
+    def on_aws?
+      if node['vagrant_environment'] == true
+        Chef::Log.info "deploying to a vagrant cluster"
```
djcp: @lbjay  Yeah, this throws out a log entry when deploying to vagrant clusters. Probably not needed any more.

libraries/default.rb
```diff
@@ -465,11 +478,13 @@ def download_episode_defaults_json_file(current_deploy_root)
         recursive true
       end
 
-      execute 'download the EpisodeDefaults.json file into the correct location' do
-        command %Q|cd #{episode_default_storage_dir} && aws s3 cp s3://#{private_assets_bucket_name}/EpisodeDefaults.json EpisodeDefaults.json|
-        retries 10
-        retry_delay 5
-        timeout 300
+      if node['vagrant_environment'] != true
```
djcp: yes, this is repeated logic. I could probably figure some way of getting rid of the repetition.

libraries/default.rb
```diff
@@ -747,10 +762,10 @@ def maven_build_for(node_profile, current_deploy_root)
       # run this in a begin/rescue block and retry a build immediately
       # after failure a few times before permanently failing
       build_profiles = {
```
djcp: This change was to support all-in-one builds. If you don't give a build profile, maven builds all jars.

recipes/install-ffmpeg.rb
```diff
@@ -10,19 +9,25 @@
...
-end
+packages = %Q|libasound2 libass4 libasyncns0 libc6 libcaca0 libdbus-1-3 libenca0 libexpat1 libflac8 libfontconfig1 libfreetype6 libfribidi0 libjson-c2 libmp3lame0 libncursesw5 libogg0 libopus0 libpng12-0 libpulse0 libsdl1.2debian libslang2 libsndfile1 libtheora0 libtinfo5 libva1 libvdpau1 libvorbis0a libvorbisenc2 libwrap0 libx11-6 libxau6 libxcb1 libxcb-shape0 libxcb-shm0 libxcb-xfixes0 libxdmcp6 libxext6 zlib1g|
+
+install_package(packages)
 
 bucket_name = get_shared_asset_bucket_name
 ffmpeg_version = node.fetch(:ffmpeg_version, '2.7.2')
 ffmpeg_archive = %Q|ffmpeg-#{ffmpeg_version}-static.tgz|
 
+if on_aws?
```
djcp: This probably needs to include the awscli installer.

PR #72, opened 2016-03-31
Add an instances_started_ok metric and alarm

djcp:
It looks at all instances in the stack - if any are in
"connection_lost", "setup_failed", or "start_failed" (which are 
non-recoverable error states), report that as an error which will get thrown
to the SNS topic for this cluster.

### DIFF
--- a/files/default/instances_started_ok_metric.sh
+++ b/files/default/instances_started_ok_metric.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+. /usr/local/bin/custom_metrics_shared.sh
+
+stack_id="$1"
+metric_name="InstancesStartedOK"
+value=0
+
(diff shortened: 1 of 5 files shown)


djcp:
I QA'd this in a cluster by intentionally breaking a chef recipe, forcing a node into "setup_failed" - this worked as expected.

files/default/instances_started_ok_metric.sh
```diff
@@ -0,0 +1,19 @@
...
+
+if [ "$output" = '' ]; then
+  # No problems
+  value=1
+else
+  # Bad things happening
+  value=0
+fi
+
+aws cloudwatch put-metric-data --region="$region" --namespace="$namespace" --dimensions="StackId=$stack_id" --metric-name="$metric_name" --value="$value"
```
lbjay: where do `$region` and `$namespace` come from?

djcp: They come from `/usr/local/bin/custom_metrics_shared.sh`

lbjay: oh ok. for some reason i didn't see that.

lbjay:
:+1:

PR #74, opened 2016-04-07
Update node

djcp:


Berksfile.lock
```diff
@@ -0,0 +1,42 @@
+DEPENDENCIES
```
djcp: This is to lock down chef recipe dependencies so we don't get any surprises. This was ignored by default by the cookbook generator, which seems very, very wrong.

recipes/install-nodejs.rb
```diff
@@ -0,0 +1,28 @@
+# Cookbook Name:: mh-opsworks-recipes
+# Recipe:: install-nodejs
+
+clear_cache = node.fetch(:clear_npm_cache, false)
```
djcp: We can force clear the npm cache through the "clear_npm_cache" variable in a stack's `custom_json` if we need to. Doesn't look like it's needed, but just in case.

recipes/install-nodejs.rb
```diff
@@ -0,0 +1,28 @@
...
+
+if clear_cache
+  execute "clear npm cache" do
+    command %Q|npm cache clear|
+    ignore_failure true
+  end
+end
+
+link '/usr/bin/node' do
```
djcp: this undoes the "legacy" system node symlink

recipes/install-nodejs.rb
```diff
@@ -0,0 +1,28 @@
...
+  action :delete
+  to '/usr/bin/nodejs'
+  ignore_failure true
+end
+
+['nodejs', 'nodejs-dev', 'npm'].each do |package_name|
```
djcp: and remove the system node packages.

recipes/install-nodejs.rb
```diff
@@ -0,0 +1,28 @@
...
+  package package_name do
+    action :purge
+    ignore_failure true
+  end
+end
+
+node.default['nodejs']['install_method'] = 'binary'
+node.default['nodejs']['version'] = node.fetch(:node_version, '0.12.1')
+include_recipe 'nodejs'
```
djcp: and now the recipe runs and installs the specific version of node that we request.

PR #77, opened 2016-04-27
MATT-2046 adding lti proxy props for mh_default_org.xml

karendolan:
```
      deploy lti config to engage & admin node deploys (not workers, they get default)
      do not need the local dns change since nginx only forwards the protocol
```

This pull is dependent on hudcede/matterhorn-dce-fork/pull-requests/187/matt-2046-enable-default-vanilla-mh-163 in bitbucket because of the new 2 arg constructor for the LtiProcessingFilter.

### DIFF
--- a/libraries/default.rb
+++ b/libraries/default.rb
@@ -657,6 +657,19 @@ module MhOpsworksRecipes
       end
     end
 
+    def install_otherpubs_service_series_impl_config(current_deploy_root)
+
+      template %Q|#{current_deploy_root}/etc/services/edu.harvard.dce.otherseries.service.OtherSeriesServiceImpl.properties| do
+        icommons_api_token = node.fetch(:icommons_api_token, 'replace-with-an-icommons-api-token-or-manually-create-series-mappings')
+        source 'edu.harvard.dce.otherseries.service.OtherSeriesServiceImpl.properties.erb'
+        owner 'matterhorn'
+        group 'matterhorn'
+        variables({
+          icommons_api_token: icommons_api_token
+        })
+      end
+    end
+
     def set_service_registry_dispatch_interval(current_deploy_root)
       template %Q|#{current_deploy_root}/etc/services/org.opencastproject.serviceregistry.impl.ServiceRegistryJpaImpl.properties| do
         source 'ServiceRegistryJpaImpl.properties.erb'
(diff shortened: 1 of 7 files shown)


libraries/default.rb
```diff
@@ -678,6 +678,29 @@ def install_multitenancy_config(current_deploy_root, admin_hostname, engage_host
       end
     end
 
+    def install_default_tentant_config(current_deploy_root, public_dns_name, private_dns_name)
```
djcp: *tenant

djcp:
Can you update the CHANGELOG with basic info about the purpose and usage of this new feature, and if it requires any explicit recipe runs to be active?

Will this work OK with a defaultly deployed development cluster?

libraries/default.rb
```diff
@@ -678,6 +678,29 @@ def install_multitenancy_config(current_deploy_root, admin_hostname, engage_host
...
+      lti_oauth_info = get_lti_auth_info
+      template %Q|#{current_deploy_root}/etc/security/mh_default_org.xml| do
+        source 'mh_default_org.xml.erb'
+        owner 'matterhorn'
+        group 'matterhorn'
+        variables({
+          lti_oauth: lti_oauth_info,
+          unproxied_name: public_dns_name,
```
djcp: At some point you think these values will be different?

recipes/deploy-admin.rb
```diff
@@ -82,6 +82,7 @@
     install_multitenancy_config(most_recent_deploy, public_admin_hostname, public_engage_hostname)
     remove_felix_fileinstall(most_recent_deploy)
     install_smtp_config(most_recent_deploy)
+    install_default_tentant_config(most_recent_deploy, public_admin_hostname, private_hostname)
```
djcp: Does it need to be active on the admin node? I'm guessing we'd only need it on nodes that run the engage-related services.

FWIW, I have no problem with configuring nodes identically even if the configs aren't used on that node type - I think it's a little clearer, actually. But if we're going to do that, then let's add this to `deploy-worker` too.

djcp:
Good work! A few questions / clarifications added for you to think about.

libraries/default.rb
spiraleddy: I initially thought I saw nginx change the DNS/IP in addition to the protocol. However, the host name was not changed when I tested on the latest deploys in my dev cluster. I left the split in here as a reminder that unproxied vs proxy host is significant, in case a forwarder changes this in the future (such as a load balancer). FYI - I hand jammed the "https" to "http" protocol change directly into the template, since I didn't know how to grab it from here.

recipes/deploy-admin.rb
spiraleddy: LTI on the admin provides alternative access to processing status, scheduling, creating series, etc. I'll fix to deploy to workers so the file is the one source for all the nodes.

spiraleddy:
FYI -  This pull is all ready to go EXCEPT that the lti_oauth consumerkey and sharedsecret params need to be added to cluster custom json config and the values need to be pulled from the secrets file.
NOTE: this has a companion pull in MH repo.

karendolan:
This pull is still pending with MH pull because the MH pull is being updated to link the existing combo pub list page. TBD

PR #81, opened 2016-05-19
matt-1875: provision recipes for flask-gunicorn apps

nmaekawa:
- create non-priviledged web user
- install python pkgs and supervisor
- virtualenv and pip requirements
- gunicorn start script virtualenv
- prevent re-create virtualenv
- redis
- logrotate for webapp related logs
- using refactor for certificates

### DIFF
--- a/recipes/configure-ca-webapp-gunicorn.rb
+++ b/recipes/configure-ca-webapp-gunicorn.rb
@@ -1,0 +1,29 @@
+# Cookbook Name:: mh-opsworks-recipes
+# Recipe:: configure-ca-webapp-gunicorn
+
+::Chef::Recipe.send(:include, MhOpsworksRecipes::RecipeHelpers)
+
+ca_webapp_info = node.fetch(:ca_webapp, {})
+app_name = ca_webapp_info.fetch(:ca_webapp_name, "webapp")
+
+execute "install gunicorn" do
+  command "source /home/web/sites/#{app_name}/venv/bin/activate && pip install gunicorn"
+  user "web"
+  creates "/home/web/sites/#{app_name}/venv/bin/gunicorn"
+end
+
+template "/home/web/sites/#{app_name}/gunicorn_start.sh" do
+  source "ca-webapp-gunicorn_start.sh.erb"
+  owner "web"
+  group "web"
+  mode "775"
+  variables({
+    ca_webapp_name: app_name
+  })
+end
+
+directory "/home/web/sock" do
+  owner "web"
+  group "web"
+  mode "775"
+end
--- a/recipes/configure-ca-webapp-nginx-proxy.rb
+++ b/recipes/configure-ca-webapp-nginx-proxy.rb
@@ -1,0 +1,27 @@
+# Cookbook Name:: mh-opsworks-recipes
+# Recipe:: configure-ca-webapp-nginx-proxy
+
+include_recipe "mh-opsworks-recipes::update-package-repo"
+::Chef::Recipe.send(:include, MhOpsworksRecipes::RecipeHelpers)
+ca_webapp_info = node.fetch(:ca_webapp, {})
+app_name = ca_webapp_info.fetch(:ca_webapp_name, "webapp")
+
+install_package("nginx")
+
+install_nginx_logrotate_customizations
+
+ssl_info = node.fetch(:ssl, get_dummy_cert)
+if cert_defined(ssl_info)
+  create_ssl_cert(ssl_info)
+  certificate_exists = true
+end
+
+template %Q|/etc/nginx/sites-enabled/default| do
+  source "ca-webapp-nginx-proxy-conf.erb"
+  manage_symlink_source true
+  variables({
+    ca_webapp: app_name
+  })
+end
+
+execute "service nginx reload"
--- a/recipes/configure-ca-webapp-supervisor.rb
+++ b/recipes/configure-ca-webapp-supervisor.rb
@@ -1,0 +1,18 @@
+# Cookbook Name:: mh-opsworks-recipes
+# Recipe:: configure-ca-webapp-supervisor
+
+include_recipe "mh-opsworks-recipes::update-package-repo"
+::Chef::Recipe.send(:include, MhOpsworksRecipes::RecipeHelpers)
+install_package("supervisor")
+
+ca_webapp_info = node.fetch(:ca_webapp, {})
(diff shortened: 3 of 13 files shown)


recipes/configure-ca-webapp-gunicorn.rb
```diff
@@ -0,0 +1,29 @@
...
+# Recipe:: configure-ca-webapp-gunicorn
+
+::Chef::Recipe.send(:include, MhOpsworksRecipes::RecipeHelpers)
+
+ca_webapp_info = node.fetch(:ca_webapp, {})
+app_name = ca_webapp_info.fetch(:ca_webapp_name, "webapp")
+
+execute "install gunicorn" do
+  command "source /home/web/sites/#{app_name}/venv/bin/activate && pip install gunicorn"
+  user "web"
```
djcp: Could we name this user something more appropriate for this tooling? maybe "capture_agent_manager?" It'd be nice to be able to understand from `ps` (and similar tools) exactly what a process is for.

recipes/configure-ca-webapp-nginx-proxy.rb
```diff
@@ -0,0 +1,27 @@
+# Cookbook Name:: mh-opsworks-recipes
+# Recipe:: configure-ca-webapp-nginx-proxy
+
+include_recipe "mh-opsworks-recipes::update-package-repo"
+::Chef::Recipe.send(:include, MhOpsworksRecipes::RecipeHelpers)
+ca_webapp_info = node.fetch(:ca_webapp, {})
+app_name = ca_webapp_info.fetch(:ca_webapp_name, "webapp")
```
djcp: instead of the very generic "webapp", could we have a more descriptive name?

recipes/configure-ca-webapp-nginx-proxy.rb
```diff
@@ -0,0 +1,27 @@
...
+
+install_nginx_logrotate_customizations
+
+ssl_info = node.fetch(:ssl, get_dummy_cert)
+if cert_defined(ssl_info)
+  create_ssl_cert(ssl_info)
+  certificate_exists = true
+end
+
+template %Q|/etc/nginx/sites-enabled/default| do
```
djcp: I built infrastructure to allow for multiple proxied backends from a single nginx frontend to exist. Let's discuss!  My concern is that this will not allow multiple backend apps to exist on a single node - maybe that's OK, but I think we were thinking this'd live on a shared utility node as it wouldn't have huge resource requirements.

recipes/install-ca-webapp-packages.rb
```diff
@@ -0,0 +1,17 @@
...
+include_recipe "mh-opsworks-recipes::update-package-repo"
+
+%w|python-dev
+python-virtualenv
+python-pip
+supervisor
+libpq-dev
+libffi-dev
+nginx|.each do |package_name|
+  install_package(package_name)
```
djcp: you can pass a string of packages to `install_package` and it'll do them all at once, which is slightly faster. Sorry for the array antipattern I used to use.

recipes/install-redis-server.rb
```diff
@@ -0,0 +1,6 @@
+# Cookbook Name:: mh-opsworks-recipes
+# Recipe:: install-redis-server
+
+::Chef::Recipe.send(:include, MhOpsworksRecipes::RecipeHelpers)
+install_package('redis-server')
```
djcp: you might want to confirm that redis is storing keys to disk periodically. I don't recall if that's the default or not.

recipes/configure-ca-webapp-supervisor.rb
```diff
@@ -0,0 +1,18 @@
...
+app_name = ca_webapp_info.fetch(:ca_webapp_name, "webapp")
+
+template %Q|/etc/supervisor/conf.d/#{app_name}.conf| do
+  source "ca-webapp-supervisor-conf.erb"
+  variables({
+    ca_webapp_name: app_name
+  })
+end
+
+execute %Q|service supervisor restart && supervisorctl start #{app_name}|
```
djcp: You probably want to confirm that supervisor is set to start on boot even if the chef recipes don't run. Easiest way to do this is to ssh into the node and run "sudo reboot", which will reboot the node without a chef recipe lifecycle events run.

recipes/configure-ca-webapp-nginx-proxy.rb
djcp: You can see this in the `recipes/configure-nginx-proxy.rb` and `recipes/update-nginx-config-for-ganglia.rb` recipes.  It might be nice to leverage this here, then we can have multiple services with frontend / backend proxies living harmoniously on this machine.

recipes/create-ca-webapp-directories.rb
```diff
@@ -0,0 +1,15 @@
+# Cookbook Name:: mh-opsworks-recipes
+# Recipe:: create-python-webapps-directories
+
+[
+  "/home/web/sites",
+  "/home/web/sock",
+  "/home/web/logs"
```
djcp: is the log rotation aggressive enough to ensure we won't fill the root partition?

djcp: Are there any other logs written besides what's in here?

recipes/install-redis-server.rb
nmaekawa: checked that the default is to save via both methods (aof, rdb). i think i'll leave this way. are you worried about disk?

recipes/configure-ca-webapp-supervisor.rb
nmaekawa: this check is just to make sure the recipe works or to be incorporated somewhere else in opsworks?

djcp: I think maybe we discussed, but it's to ensure that supervisor starts correctly at boot. It should, but this should be something we confirm for any new services we install.

recipes/install-redis-server.rb
djcp: no, just persistence. You don't think this'll ever get large, right? Do you have a way to expire keys? IIRC, you can do this directly with redis.

nmaekawa: ok, i'm expiring keys every 24h. if i ended up throwing more info in redis i'll review persistence and expiration, but for now it should not grow too much. it mainly has a fixed set of capture agents (that does not grow during a semester) and user sessions - and this app is for internal use, so the number of users should be limited.

PR #83, opened 2016-05-24
matt-1875: provision recipes for flask-gunicorn apps

nmaekawa:
- create non-priviledged user
- install python pkgs and supervisor
- virtualenv and pip requirements
- gunicorn start script virtualenv
- prevent re-create virtualenv
- redis
- logrotate for app related logs
- using refactor for certificates
- (fix pull #81) more descriptive names
- (fix pull #81) string of packages to `install-packages` recipe
- (fix pull #81) redis default save to disk is enough for now
- (fix pull #81) nginx configs
- (fix pull #81) logrotate keeps 7 logs instead of 45

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,15 @@
 
 ## TO BE RELEASED
 
-* recipes to install and configure the utility node to host the capture-agent-manager
-  flask-gunicorn webapp
+* *[OPTIONAL]* Recipes to install and configure the utility node to host the
+  capture-agent-manager flask-gunicorn webapp. The utility node is optional and
+  only for DCE production cluster.
 
-* *REQUIRES EDITS TO THE CLUSTER CONFIG* *REQUIRES MANUAL CHEF RECIPE RUNS*:
+* *[OPTIONAL] EDITS TO THE CLUSTER CONFIG* *REQUIRES MANUAL CHEF RECIPE RUNS*:
   recipes, etc to install and configure the utility node to host the capture-agent
-  manager flask-gunicorn webapp. see README.capture-agent-manager.md in
+  manager flask-gunicorn webapp. See README.capture-agent-manager.md in
   mh-opsworks for setup instructions.
+
   The manual recipe run can be executed any time since they affect only the
   utility node
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 ## TO BE RELEASED
 
+* recipes to install and configure the utility node to host the capture-agent-manager
+  flask-gunicorn webapp
+
+* *REQUIRES EDITS TO THE CLUSTER CONFIG* *REQUIRES MANUAL CHEF RECIPE RUNS*:
+  recipes, etc to install and configure the utility node to host the capture-agent
+  manager flask-gunicorn webapp. see README.capture-agent-manager.md in
+  mh-opsworks for setup instructions.
+  The manual recipe run can be executed any time since they affect only the
+  utility node
+
+        ./bin/rake stack:commands:execute_recipes_on_layers layers="Utility"
+        recipes="mh-opsworks-recipes::create-capture-agent-manager-user,mh-opsworks-recipes::create-capture-agent-manager-directories,mh-opsworks-recipes:install-capture-agent-manager-packages,mh-opsworks-recipes:install-capture-agent-manager,mh-opsworks-recipes:configure-capture-agent-manager-gunicorn,mh-opsworks-recipes:configure-capture-agent-manager-nginx-proxy,mh-opsworks-recipes:configure-capture-agent-manager-supervisor"
+
+
 ## 1.3.3 - 5/19/2016
 
 * Fix all-in-one node provisioning, primarily for local vagrant development.
--- a/libraries/default.rb
+++ b/libraries/default.rb
@@ -382,6 +382,27 @@ module MhOpsworksRecipes
         mode '0600'
       end
     end
+
+    def get_capture_agent_manager_info
+      node.fetch(
+        :capture_agent_manager, {
+          ca_stats_user: 'user',
+          ca_stats_passwd: 'passwd',
+          ca_stats_json_url: 'http://ca-status.dceapp.net/ca_stats/ca_stats.json',
+          epipearl_user: 'admin',
+          epipearl_passwd: 'passwd',
+          ldap_host: 'dev-ldap1.dce.harvard.edu',
+          ldap_base_search: 'dc=dce,dc=harvard,dc=edu',
+          ldap_bind_dn: 'cn=user,dc=dce,dc=harvard,dc=edu',
+          ldap_bind_passwd: 'passwd',
+          capture_agent_manager_secret_key: 'super_secret_really',
+          log_config: 'logging.yaml',
+          memcached_port: '8008',
+          capture_agent_manager_git_repo: 'https://github.com/harvard-dce/capture_agent_manager',
+          capture_agent_manager_git_revision: 'master'
+          }
+        )
+    end
   end
 
   module DeployHelpers
--- a/libraries/default.rb
+++ b/libraries/default.rb
@@ -389,6 +389,7 @@ module MhOpsworksRecipes
           capture_agent_manager_app_name: 'cadash',
           capture_agent_manager_usr_name: 'capture_agent_manager',
           capture_agent_manager_secret_key: 'super_secret_key',
+          capture_agent_manager_gunicorn_log_level: 'debug',
           log_config: '/home/capture_agent_manager/sites/cadash/logging.yaml',
           ca_stats_user: 'usr',
           ca_stats_passwd: 'pwd',
@@ -399,7 +400,7 @@ module MhOpsworksRecipes
           ldap_base_search: 'dc=some-domain,dc=com',
           ldap_bind_dn: 'cn=fake_usr,dc=some-domain,dc=com',
           ldap_bind_passwd: 'pwd',
-          capture_agent_manager_git_repo: 'https://github.com/capture_agent_git_repo',
+          capture_agent_manager_git_repo: 'https://github.com/harvard-dce/cadash',
           capture_agent_manager_git_revision: 'master',
           capture_agent_manager_database_usr: 'usr',
           capture_agent_manager_database_pwd: 'pwd'
--- a/libraries/default.rb
+++ b/libraries/default.rb
@@ -382,6 +382,40 @@ module MhOpsworksRecipes
         mode '0600'
       end
     end
+
+    def get_capture_agent_manager_info
+      node.fetch(
+        :capture_agent_manager, {
+          capture_agent_manager_app_name: 'cadash',
+          capture_agent_manager_usr_name: 'capture_agent_manager',
+          capture_agent_manager_secret_key: 'super_secret_key',
+          log_config: '/home/capture_agent_manager/sites/cadash/logging.yaml',
+          ca_stats_user: 'usr',
+          ca_stats_passwd: 'pwd',
+          ca_stats_json_url: 'http://fake-ca-status.com/ca-status.json',
+          epipearl_user: 'usr',
+          epipearl_passwd: 'pwd',
+          ldap_host: 'ldap-hostname.some-domain.com',
+          ldap_base_search: 'dc=some-domain,dc=com',
+          ldap_bind_dn: 'cn=fake_usr,dc=some-domain,dc=com',
+          ldap_bind_passwd: 'pwd',
+          capture_agent_manager_git_repo: 'https://github.com/capture_agent_git_repo',
+          capture_agent_manager_git_revision: 'master',
+          capture_agent_manager_database_usr: 'usr',
+          capture_agent_manager_database_pwd: 'pwd'
+        }
+      )
+    end
+
+    def get_capture_agent_manager_app_name
+      ca_info = get_capture_agent_manager_info
+      ca_info[:capture_agent_manager_app_name]
+    end
+
+    def get_capture_agent_manager_usr_name
+      ca_info = get_capture_agent_manager_info
+      ca_info[:capture_agent_manager_usr_name]
+    end
   end
 
   module DeployHelpers
--- a/libraries/default.rb
+++ b/libraries/default.rb
@@ -382,27 +382,6 @@ module MhOpsworksRecipes
         mode '0600'
       end
     end
-
-    def get_capture_agent_manager_info
-      node.fetch(
-        :capture_agent_manager, {
-          ca_stats_user: 'user',
-          ca_stats_passwd: 'passwd',
-          ca_stats_json_url: 'http://ca-status.dceapp.net/ca_stats/ca_stats.json',
-          epipearl_user: 'admin',
-          epipearl_passwd: 'passwd',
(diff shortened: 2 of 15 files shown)


libraries/default.rb
```diff
@@ -382,6 +382,27 @@ def create_ssl_cert(ssl_info)
         mode '0600'
       end
     end
+
+    def get_capture_agent_manager_info
```
djcp: is this used anywhere?

recipes/configure-capture-agent-manager-gunicorn.rb
```diff
@@ -0,0 +1,30 @@
+# Cookbook Name:: mh-opsworks-recipes
+# Recipe:: configure-capture-agent-manager-gunicorn
+
+::Chef::Recipe.send(:include, MhOpsworksRecipes::RecipeHelpers)
+
+capture_agent_manager_info = node.fetch(:capture_agent_manager, {})
+app_name = capture_agent_manager_info.fetch(:capture_agent_manager_name, "capture_agent_manager")
+
+execute "install gunicorn" do
+  command "source /home/capture_agent_manager/sites/#{app_name}/venv/bin/activate" \
```
djcp: Oh! This works! Neat. I've never thought to do multiline strings like this!

recipes/configure-capture-agent-manager-gunicorn.rb
```diff
@@ -0,0 +1,30 @@
...
+  source "capture-agent-manager-gunicorn-start.sh.erb"
+  owner "capture_agent_manager"
+  group "capture_agent_manager"
+  mode "775"
+  variables({
+    capture_agent_manager_name: app_name
+  })
+end
+
+directory "/home/capture_agent_manager/sock" do
```
djcp: should this be created before the other stuff is run?

recipes/configure-capture-agent-manager-supervisor.rb
```diff
@@ -0,0 +1,18 @@
...
+app_name = capture_agent_manager_info.fetch(:capture_agent_manager_name, "capture_agent_manager")
+
+template %Q|/etc/supervisor/conf.d/#{app_name}.conf| do
+  source "capture-agent-manager-supervisor-conf.erb"
+  variables({
+    capture_agent_manager_name: app_name
+  })
+end
+
+execute %Q|service supervisor restart && supervisorctl start #{app_name}|
```
djcp: the restart doesn't automatically start the `app_name` ?  But this still works correctly when you do a `sudo reboot`?

templates/default/capture-agent-manager-gunicorn-start.sh.erb
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+NAME="<%= @capture_agent_manager_name %>"
+FLASKDIR="/home/capture_agent_manager/sites/${NAME}"
+VENVDIR="${FLASKDIR}/venv"
+SOCKFILE="/home/capture_agent_manager/sock/${NAME}.sock"
+USER=capture_agent_manager
```
djcp: should we template this name?

djcp:
looks good! There are a few places where you could probably template the name a little more thoroughly.

Where are you thinking you'd document the purpose of this an how to use it? If you follow the pattern @lbjay and I have been following, it'd be in a README in mh-opsworks proper.

libraries/default.rb
nmaekawa: i went back and forth but i'm using this now (indirectly) on most capture-agent-manager recipes.

recipes/configure-capture-agent-manager-gunicorn.rb
nmaekawa: removed this since these dirs are created in a recipe that is run earlier.

recipes/configure-capture-agent-manager-supervisor.rb
nmaekawa: i think i wasn't sure and run both just in case. i'm sure now since i've tested this a couple of times and it works how is supposed to. removed the chained `supervisor start <app>`.

nmaekawa:
added info on metadata.rb and CHANGELOG.md.
also, lots of little edits to distinguish the capture-agent-manager username from capture-agent-manager app name, and to make it `configurable` -- i think when i started writing the recipes the app would be `redunlive` and i changed to `cadash` to include other related functionalities (and wanted the app name to be configurable because of that)

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
 
 ## TO BE RELEASED
 
+* recipes to install and configure the utility node to host the capture-agent-manager
+  flask-gunicorn webapp
+
+* *REQUIRES EDITS TO THE CLUSTER CONFIG* *REQUIRES MANUAL CHEF RECIPE RUNS*:
```
djcp: I would update this language to make sure it's understood this is optional and only for the DCE production cluster.

libraries/default.rb
```diff
@@ -382,6 +382,40 @@ def create_ssl_cert(ssl_info)
...
+        }
+      )
+    end
+
+    def get_capture_agent_manager_app_name
+      ca_info = get_capture_agent_manager_info
+      ca_info[:capture_agent_manager_app_name]
+    end
+
+    def get_capture_agent_manager_usr_name
```
djcp: well, "e" is one of the most common letters in the english language - I guess we're running out? :)

metadata.rb
```diff
@@ -867,6 +867,101 @@
...
 '
 )
+recipe(
+  'mh-opsworks-recipes::configure-capture-agent-manager-gunicorn',
+  'sets up start script for gunicorn with capture-agent-manager app
+
+this is relevant for the utility node, where the capture-agent-manager `cadash`
+should run.
+
+=== attributes
```
djcp: nice! but for consistency it'd be nice to match capitalization of all the other headers (nitpicky, yes).

recipes/configure-capture-agent-manager-gunicorn.rb
```diff
@@ -0,0 +1,25 @@
+# Cookbook Name:: mh-opsworks-recipes
+# Recipe:: configure-capture-agent-manager-gunicorn
+
+::Chef::Recipe.send(:include, MhOpsworksRecipes::RecipeHelpers)
+
+app_name = get_capture_agent_manager_app_name
+username = get_capture_agent_manager_usr_name
```
djcp: "usr_name" and "username". . .

djcp:
A couple relatively minor clarifications, otherwise looks good.

libraries/default.rb
```diff
@@ -382,6 +382,40 @@ def create_ssl_cert(ssl_info)
...
+          ca_stats_user: 'usr',
+          ca_stats_passwd: 'pwd',
+          ca_stats_json_url: 'http://fake-ca-status.com/ca-status.json',
+          epipearl_user: 'usr',
+          epipearl_passwd: 'pwd',
+          ldap_host: 'ldap-hostname.some-domain.com',
+          ldap_base_search: 'dc=some-domain,dc=com',
+          ldap_bind_dn: 'cn=fake_usr,dc=some-domain,dc=com',
+          ldap_bind_passwd: 'pwd',
+          capture_agent_manager_git_repo: 'https://github.com/capture_agent_git_repo',
```
lbjay: i'm not sure it's necessary to make this configurable. will it change? if you look at the `install-ua-harvester` recipe I just put the repo path right in the code.

recipes/configure-capture-agent-manager-gunicorn.rb
```diff
@@ -0,0 +1,25 @@
+# Cookbook Name:: mh-opsworks-recipes
+# Recipe:: configure-capture-agent-manager-gunicorn
+
+::Chef::Recipe.send(:include, MhOpsworksRecipes::RecipeHelpers)
+
+app_name = get_capture_agent_manager_app_name
+usr_name = get_capture_agent_manager_usr_name
+
+execute "install gunicorn" do
+  command "source /home/#{usr_name}/sites/#{app_name}/venv/bin/activate" \
```
lbjay: fyi, not a biggie but you can run `/home/#{usr_name}/sites/#{app_name}/venv/bin/pip` directly and it'll work in the context of that virtualenv without the `source` part

lbjay: Also, is gunicorn not installed via the cadash's requirements.txt?

recipes/configure-capture-agent-manager-nginx-proxy.rb
```diff
@@ -0,0 +1,38 @@
...
+usr_name = get_capture_agent_manager_usr_name
+
+install_package("nginx")
+
+install_nginx_logrotate_customizations
+
+ssl_info = node.fetch(:ca_ssl, get_dummy_cert)
+if cert_defined(ssl_info)
+  create_ssl_cert(ssl_info)
+  certificate_exists = true
```
lbjay: It doesn't look like you're using `certificate_exists` anywhere, which makes me wonder if it would ever be the case that ssl would not be enabled? If not you might consider simplifying, e.g., the `install-kibana` recipe, where I just include the dummy cert in the default configuration.

templates/default/capture-agent-manager-supervisor-conf.erb
```diff
@@ -0,0 +1,14 @@
+[program:<%= @capture_agent_manager_name %>]
+command = /home/<%= @capture_agent_manager_usr_name %>/sites/<%= @capture_agent_manager_name %>/gunicorn_start.sh
+directory = /home/<%= @capture_agent_manager_usr_name %>/sites/<%= @capture_agent_manager_name %>
+user = <%= @capture_agent_manager_usr_name %>
+redirect_stderr = True
+
+stdout_logfile = /home/<%= @capture_agent_manager_usr_name %>/logs/<%= @capture_agent_manager_name %>_gunicorn_stdout.log
+stderr_logfile = /home/<%= @capture_agent_manager_usr_name %>/logs/<%= @capture_agent_manager_name %>_gunicorn_stderr.log
```
lbjay: I don't think you need this since you've got `redirect_stderr = True`. Everything is going to go to `stdout_logfile`.

templates/default/nginx-proxy-capture-agent-manager.conf.erb
```diff
@@ -0,0 +1,10 @@
+    location / {
+        proxy_buffering off;
+        proxy_pass http://unix:/home/<%= @capture_agent_manager_usr_name %>/sock/<%= @capture_agent_manager %>.sock;
```
lbjay: is there a reason you chose sockets vs tcp here?

templates/default/nginx-proxy-ssl-only.erb
```diff
@@ -0,0 +1,33 @@
+server {
+    listen 443 default_server ssl;
```
lbjay: It's pretty simple to listen on :80 + redirect to https too. See `elk-nginx-proxy-conf.erb`

templates/default/capture-agent-manager-gunicorn-start.sh.erb
```diff
@@ -0,0 +1,32 @@
...
+
+echo "starting ${NAME}"
+
+# activate virtualenv
+source ${VENVDIR}/bin/activate
+
+export PYTHONPATH=${FLASKIDR}:${PYTHONPATH}
+
+# export all env vars for app
+source ${FLASKDIR}/${NAME}.env
```
lbjay: I'm not used to seeing `.env` files used in this way, but I think it's probably OK if makes things easier. 

Normally I would expect the `.env` to just list the `VARIABLE=foo` pairs (without the `export`), then your flask app would read/use the settings via something like https://github.com/grauwoelfchen/flask-dotenv

recipes/configure-capture-agent-manager-supervisor.rb
```diff
@@ -0,0 +1,19 @@
+# Cookbook Name:: mh-opsworks-recipes
+# Recipe:: configure-capture-agent-manager-supervisor
+
+include_recipe "mh-opsworks-recipes::update-package-repo"
+::Chef::Recipe.send(:include, MhOpsworksRecipes::RecipeHelpers)
+install_package("supervisor")
```
lbjay: I :heart: supervisord

recipes/install-capture-agent-manager.rb
```diff
@@ -0,0 +1,55 @@
...
+end
+
+execute "create virtualenv" do
+  command %Q|/usr/bin/virtualenv /home/#{usr_name}/sites/#{app_name}/venv|
+  user usr_name
+  creates %Q|/home/#{usr_name}/sites/#{app_name}/venv/bin/activate|
+end
+
+execute "install capture_agent_manager dependencies" do
+  command %Q|source /home/#{usr_name}/sites/#{app_name}/venv/bin/activate && pip install -r /home/#{usr_name}/sites/#{app_name}/requirements.txt|
```
lbjay: Again, you should be able to skip the `source`ing and just exec `/home/#{usr_name}/sites/#{app_name}/venv/bin/pip` directly.

templates/default/capture-agent-manager-gunicorn-start.sh.erb
```diff
@@ -0,0 +1,32 @@
...
+# create the run directory if it doesn't exist
+RUNDIR=$(dirname ${SOCKFILE})
+test -d ${RUNDIR} || mkdir -p ${RUNDIR}
+
+# start gunicorn
+exec gunicorn ${NAME}.app:create_app\(\) \
+    --name ${NAME} \
+    --workers ${NUM_WORKERS} \
+    --user=${USER} --group=${GROUP} \
+    --log-level=debug \
```
lbjay: should this be a config setting in `get_capture_agent_manager_info`?

libraries/default.rb
nmaekawa: hum.. this is because i was working on a fork.

recipes/configure-capture-agent-manager-gunicorn.rb
nmaekawa: yes, it does. how much re-usable a recipe should be? i could write this assuming that the app requirements already run and just install the gunicorn-start script?

recipes/configure-capture-agent-manager-nginx-proxy.rb
nmaekawa: this app requires ssl because of ldap login... but i think i left this way so when in production i can set a real certificate.

templates/default/nginx-proxy-capture-agent-manager.conf.erb
nmaekawa: i copied from an example i googled... but looks nice that it will not allow anyone to access the app via some tcp port. do you think this has drawbacks?

templates/default/nginx-proxy-ssl-only.erb
nmaekawa: i was doing that but then dan suggested that we not even redirecting.

templates/default/capture-agent-manager-gunicorn-start.sh.erb
nmaekawa: i think it used to be just variable=foo when it was set up for heroku... what should be extension then? `.sh`? i guess i'm still learning these patterns to setup stuff in the env.

lbjay:
:+1:

PR #86, opened 2016-06-10
bucket for storing last action timestamp needs to be unique per cluster

lbjay:
I'd been using a single bucket with s3 objects differentiated by cluster name. This was backwards (duh) as you can't re-use a bucket across accounts (dce-deac vs dce-deac-test).


recipes/install-ua-harvester.rb
```diff
@@ -65,7 +65,7 @@
 cron_d 'ua_harvester' do
   user 'ua_harvester'
   minute '*/2'
-  command %Q(cd /home/ua_harvester/harvester && /usr/bin/run-one ./ua_harvest.py 2>&1 | logger -t info)
+  command %Q(cd /home/ua_harvester/harvester && /usr/bin/run-one ./ua_harvest.py -b 10000 2>&1 | logger -t info)
```
lbjay: oops. this is an unrelated change. just bumping the batch size from the default 1000/fetch.

PR #88, opened 2016-06-21
Mi 46 newrelic

phoebem:
[MI-46-newrelic] Updated newrelic configuration to be specific to layer type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,16 +2,17 @@
 
 ## TO BE RELEASED
 
-
-## 1.5.1 - 6/20/2016
 * Configuration for the newrelic is now different for each layer (admin,work,engage)
-"custom_json": {
-  "newrelic": {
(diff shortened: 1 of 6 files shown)


lbjay:
To squash the commits into one do:

`git rebase -i HEAD~n` where `n` is the number of commits. (currently 11)

then in the edit interface that appears replace "pick" with "squash" for all but the 1st commit. Then exit from the editor. 

You'll then need to force push to the remote branch:
`git push --force origin MI-46-newrelic`

CHANGELOG.md
```diff
@@ -3,6 +3,16 @@
 ## TO BE RELEASED
 
 
+## 1.5.1 - 6/20/2016
```
lbjay: Ooops, sorry for missing this. You need to move the release notes into the "TO BE RELEASED" section. Also the section should include notes on what/how recipes are to be run at deploy time.

PR #93, opened 2016-06-30
MATT-2103-two-week-pending-tasks Added stack name to config.properties.

rute-santos:
This was added to allow showing the cluster name in the subject of notification emails.

### DIFF
--- a/recipes/deploy-admin.rb
+++ b/recipes/deploy-admin.rb
@@ -10,6 +10,7 @@ storage_info = get_storage_info
 shared_storage_root = get_shared_storage_root
 rest_auth_info = get_rest_auth_info
 admin_user_info = get_admin_user_info
+stack_name = stack_shortname
 
 capture_agent_query_url = node.fetch(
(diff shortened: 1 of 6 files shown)


recipes/deploy-admin.rb
```diff
@@ -10,6 +10,7 @@
 shared_storage_root = get_shared_storage_root
 rest_auth_info = get_rest_auth_info
 admin_user_info = get_admin_user_info
+stack_name = stack_shortname
```
spiraleddy: Rute, where does the stack_shortname come from? Does it come from the cluster config's stack.name? 
ie. 
{
  "version": 102,
  [...]
   "stack": {
    "name": "kdolan-dev",

rute-santos:
Yes, with some characters changed to - (in libraries/default.rb):
    def stack_shortname
      stack_name = node[:opsworks][:stack][:name].gsub(/[^a-z\d\-]/,'-')
    end

PR #97, opened 2016-08-08
mo-scaler updates + queued jobs metric

lbjay:
Lumping these commits together as they are closely related.

The mo-scaler configuration has changed dramatically to allow monitoring of multiple cloudwatch metrics as well as other sources of info. Therefore the configuration deployment needed to change. Gone are most of the `AUTOSCALE_...` env vars, replaced with a json config file which is generated by the recipe based on the `moscaler` settings in the cluster config. 

Also, I've resuscitated and updated a recipe that generates a `MatterhornJobsQueeu` cloudwatch metric. This will hopefully be the (or one of the) metrics that the horizontal scaling takes into consideration for scaling up/down.

### DIFF
--- a/files/default/queued_job_count.py
+++ b/files/default/queued_job_count.py
@@ -18,19 +18,22 @@ def main(args):
     # then get their running operations
     running_ops = []
     for wf in running_wfs:
-        running_ops.extend(filter(
-            lambda x: x.state in ["RUNNING","WAITING"],
-            wf.operations
-        ))
+        running_ops.extend([
+            x for x in wf.operations
+            if x.state in ["RUNNING", "WAITING"]
+        ])
 
     # filter for the operation types we're interested in
-    if args.type is not None:
-        running_ops = filter(lambda x: x.id in args.type, running_ops)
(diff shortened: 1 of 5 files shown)


files/default/queued_job_count.py
```diff
@@ -18,19 +18,22 @@ def main(args):
     # then get their running operations
     running_ops = []
     for wf in running_wfs:
-        running_ops.extend(filter(
-            lambda x: x.state in ["RUNNING","WAITING"],
-            wf.operations
-        ))
+        running_ops.extend([
+            x for x in wf.operations
+            if x.state in ["RUNNING", "WAITING"]
```
rute-santos: Where does WAITING come from? Is this related to Greg's future job weights code?

lbjay: Correct. It's been in there for awhile in anticipation of the job weights code getting merged someday.

lbjay:
Ugh. I somehow managed to do this pull request against master (although my dev branch was based off develop). Changes have been merged to develop. Closing manually.

PR #99, opened 2016-09-02
Allow autofs restart if `storage_hostname` differs from current mount

lbjay:
This allows the `nfs-client` recipe to restart autofs in the situation where the storage hostname has changed. This would happen in the case of a storage node or VPSA swap.


lbjay:
Successfully tested in my `jluker-zadara-switch` cluster

nmaekawa:
looks good to me.

PR #100, opened 2016-09-02
cam-20-timedrift: install cronjob to poll time in CAs

nmaekawa:
this needs documentation, but it goes in the mh-opsworks repo
https://github.com/harvard-dce/mh-opsworks/pull/103

also, for the ssh command in the template, i've set to "StrictHostKeyChecking no". this is compromises security but is needed when we add a new CA, cause ssh is going to interactively ask if it's ok to add new host (and it's the first time running in a fresh installation).

### DIFF
--- a/templates/default/ca-timedrift-metric-script.erb
+++ b/templates/default/ca-timedrift-metric-script.erb
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+namespace="DECustom"
+region="us-east-1"
+metric_name="CA-timedrift"
+dimension_name="CAName"
+
+ca_list=$(curl -v --user "<%= @ca_stats_user %>:<%= @ca_stats_passwd %>" "<%= @ca_stats_json_url %>" | grep address | awk  -F'"' '{print $4;}' | awk -F. '{print $1;}')
+
+for ca in ${ca_list}
+do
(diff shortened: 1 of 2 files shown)


templates/default/ca-timedrift-metric-script.erb
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+namespace="DECustom"
+region="us-east-1"
+metric_name="CA-timedrift"
+dimension_name="CAName"
+
+ca_list=$(curl -v --user "<%= @ca_stats_user %>:<%= @ca_stats_passwd %>" "<%= @ca_stats_json_url %>" | grep address | awk  -F'"' '{print $4;}' | awk -F. '{print $1;}')
+
+for ca in ${ca_list}
```
lbjay: how many capture agents are we looping through here? I'm concerned about API throttling with this many rapid calls. there are methods by which you can bundle the metric data into a json file and publish with one call, but it would introduce some complexity. Was this tested with the full list of CAs?

lbjay: Hmmm, although I just realized there's probably a delay between calls due to the ssh calls in the loop. Maybe you could add an extra `sleep 1` in the loop to be safe?

nmaekawa: this runs with the complete list of CAs, including devs.
it's been running during the weekend in the test account.

PR #103, opened 2016-09-28
ELK component version updates + new episode index creation.

lbjay:
Includes:
* explicitly install openjdk
* use strings vs symbols to access elk_info hash
* bump default versions of packages & "pin" major versions
* install kibana via apt repo
* add index template for episodes
* add config & cron entry for indexing episode data
* updated ua harvester `.env` settings

See my comments in the diff for additional details on some of these things. Also there will be a corresponding PR for harvard-dce/mh-user-action-harvester.

### DIFF
--- a/libraries/default.rb
+++ b/libraries/default.rb
@@ -36,6 +36,13 @@ module MhOpsworksRecipes
       end
     end
 
+    def pin_package(name, version)
+      apt_preference name do
+        pin "version #{version}"
+        pin_priority "700"
+      end
+    end
+
     def get_shared_asset_bucket_name
       node.fetch(:shared_asset_bucket_name, 'mh-opsworks-shared-assets')
     end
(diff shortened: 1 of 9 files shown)


libraries/default.rb
```diff
@@ -329,23 +336,25 @@ def git_repo_url(git_data)
 
     def get_elk_info
       stack_name = stack_shortname
-      ::Chef::Mixin::DeepMerge.deep_merge({
```
lbjay: Reworked many of the version settings here and added apt repo urls to facilitate easier pinning. In addition, I switched to using string hash keys (vs symbols) after discovering that the merging of default values and cluster config overrides was actually working (!).

libraries/default.rb
```diff
@@ -36,6 +36,13 @@ def install_package(name)
       end
     end
 
+    def pin_package(name, version)
```
lbjay: This is from the [apt cookbook](https://supermarket.chef.io/cookbooks/apt). It creates files in /etc/apt/preferences.d/ that restrict which versions of a package apt is allowed to install or update.

For example, we want to pin elasticsearch to the major version `2.4` so the `install-elasticsearch` recipe calls `pin_package("elasticsearch", "2.4.*")` which results in a file `/etc/apt/preferences.d/elasticsearch.pref` that contains:

    Package: elasticsearch
    Pin: version 2.4.*
    Pin-Priority: 700

Currently the latest release of elasticsearch is 4.6.1. Should 4.6.2 be released an update packages operation would update the package, whereas a 4.7.x release would not be allowed, i.e. we would only get patch-level automatic updates.

recipes/install-elasticsearch.rb
```diff
@@ -6,35 +6,39 @@
...
+es_cluster_name = elk_info['es_cluster_name']
+data_path = elk_info['es_data_path']
 es_heap_size = xmx_ram_for_this_node(0.5)
 es_host = node[:opsworks][:instance][:private_ip]
 region = node[:opsworks][:instance][:region]
 stack_name = stack_shortname
-enable_snapshots = elk_info[:es_enable_snapshots]
-es_repo_bucket = "#{stack_name}-snapshots"
+enable_snapshots = elk_info['es_enable_snapshots']
+es_snapshot_bucket = "#{stack_name}-snapshots"
```
lbjay: just renaming a variable here to be more clear about what it represents

recipes/install-ua-harvester.rb
```diff
@@ -46,8 +48,10 @@
 AWS_DEFAULT_REGION="#{region}"
 MATTERHORN_REST_USER="#{rest_auth_info[:user]}"
 MATTERHORN_REST_PASS="#{rest_auth_info[:pass]}"
-MATTERHORN_HOST="#{engage_node[:private_ip]}"
-S3_LAST_ACTION_TS_BUCKET="#{stack_name}-ua-harvester"
+MATTERHORN_ENGAGE_HOST="#{engage_node[:private_ip]}"
```
lbjay: The `load_episodes` process needs to use pyhorn clients pointed at both engage and admin both.

lbjay:
Upgrades to the pipeline components have passed QA: see QA-418

recipes/install-kibana.rb
```diff
@@ -11,49 +11,24 @@
...
-create_ssl_cert(elk_info[:http_ssl])
+create_ssl_cert(elk_info['http_ssl'])
 
-group 'kibana' do
-  append true
+apt_repository 'kibana' do
+  uri kibana_repo_uri
+  components ['stable', 'main']
+  keyserver 'ha.pool.sks-keyservers.net'
+  key '46095ACC8548582C1A2699A9D27D666CD88E42B4'
```
nmaekawa: these keys are not secrets? they can go into the repo without problems?

lbjay: right. they're public keys for the package repos. needed when the underlying chef code runs apt-add-repository.

PR #104, opened 2016-10-05
jvm metrics to cloudwatch

lbjay:


lbjay:
In the dev account you can see the generated metrics in the Custom Metrics -> Java/felix namespace.

lbjay:
Once again based the PR off master instead of develop and now that i've merged into develop the PR won't let me change the base. Closed.

PR #106, opened 2016-10-07
MATT-2173 remove submodule init

karendolan:
This removes the submodule init from the deploy. Paella player module source files are now directly available in the module and no longer retrieved from a submodule. (The source files are pre-compiled from all its dependencies during the dev phase).


karendolan:
NOTE: This pull is dependent on the MH, DCE paella-matterhorn & dce-paella-extensions pulls. They all need to go together.

PR #108, opened 2016-10-12
recipe and configs for the cloudwatch log agent

lbjay:


karendolan:
Hi Jay, the AWS cloudwatch logs is going on 7 logs? Do we have an idea of what that  will consumer for resources (https://forums.aws.amazon.com/thread.jspa?messageID=605516) ?

lbjay:
Good catch. I hadn't considered that it would be significant, but I'll look into it. FWIW, the number of log readers would be highest on admin at 5 total (nginx access/error, matterhorn, syslog and mail.log). I threw mail.log in there for kicks since it might be interesting to track notifications to producers, but it's probably not necessary.

lbjay:
@karendolan htop on my admin node (5 logs being streamed) showed the combined CPU% as 0.5-1.5% which seems negligible but this was with very little activity. I think it's something I'll need to monitor when it hits staging.

karendolan:
@lbjay per that article, the CPU load is related to the amount of logging in the watched logs.

recipes/set-timezone.rb
```diff
@@ -3,8 +3,14 @@
...
+# need to create a service stub to notify
+service 'rsyslog' do
+  action :nothing
+end
+
 execute 'set timezone' do
   command %Q|timedatectl set-timezone "#{timezone}"|
   retries 5
   retry_delay 5
+  notifies :restart, "service[rsyslog]", :immediately
```
lbjay: without this the syslog timestamps keep using the old timezone (which makes the cloudwatch log agent complain & fail)

PR #110, opened 2016-10-24
redis for harvester caching + analytics node recipe refactoring

lbjay:
includes:
* redis install for havester episode caching
* separating ua-havester install from configure
* fixing use of chef `service` and `notifies` directives
* refatoring and consolidation of logstash/kibana installation

The adding of redis is just one line change to install the package.

Separating the ua-harvester install vs configure is an overdue change. The harvester config needs to be re-generated on a cluster 'configure' lifecycle event as it needs to get the engage/admin IP should they change for some reason. But the current version of the recipe is doing way more than that. I broke it up into `install-ua-harvester` and `configure-ua-harvester`.

The rest of the changes to the logstash/kibana stuff is me cleaning up some chef cluelessness from my initial implementation.


files/default/kibana-4.x-default
```diff
@@ -1,6 +0,0 @@
-user="kibana"
```
lbjay: this file isn't necessary anymore since we switched to installing kibana from a .deb package on a recent release

files/default/kibana-4.x-init
```diff
@@ -1,157 +0,0 @@
-#!/bin/sh
```
lbjay: same with this file. the kibana .deb package installs it's own service init script.

libraries/default.rb
```diff
@@ -475,6 +475,10 @@ def configure_cloudwatch_log(log_name, log_file, datetime_format)
       retention_days = node.fetch(:cwlogs_retention_days, '30')
       region = node.fetch(:region, 'us-east-1')
 
+      service 'awslogs' do
```
lbjay: without this the `template` block further down doesn't know how to notify the awslogs service to restart. This is OK when called in a  chef run that includes the `install-cwlogs` recipe, but fails in other contexts, e.g., a one-off run of a recipe that tries to use `configure_cloudwatch_log`.

recipes/configure-logstash-kibana.rb
```diff
@@ -1,38 +0,0 @@
-# Cookbook Name:: mh-opsworks-recipes
```
lbjay: this functionality was split up and added to the respecting logstash/kibana install recipes

karendolan:
@lbjay How easy is it to separate the UA harvester from the MH deployment recipes? Could I deploy just an independent UA harvester on a DCE AWS node and point it to what ever MH cluster I wanted (with proper permissions)? Or does it have to be deployed into an existing MH cluster?

lbjay:
@karendolan good question. It can be done! I'll add a section to the readme to explain how one would set things up independent of a cluster/chef/etc.

lbjay:
@karendolan I've added another commit to the PR at https://github.com/harvard-dce/mh-user-action-harvester/pull/3 that includes additional setup instructions and examples in the README.

PR #112, opened 2016-10-31
cadash db provisioning

lbjay:
*DO NOT MERGE*


files/default/capture-agent-manager-logrotate.conf
```diff
@@ -1,8 +0,0 @@
-/home/capture_agent_manager/logs/*.log {
-	daily
-	missingok
-	rotate 7
-	compress
-	notifempty
-	copytruncate
-}
```
lbjay: Removing as I think we should just log to syslog (which requires changes to cadash `logging.yml`)

templates/default/capture-agent-manager-gunicorn-upstart.conf.erb
```diff
@@ -0,0 +1,19 @@
+description "Gunicorn for capture-agent-manager"
+
+start on (filesystem)
+stop on runlevel [016]
+respawn
+respawn limit 10 5
+
+setuid <%= @user %>
```
lbjay: /home/capture_agent_manager/sites/cadash/gunicorn_config.py

templates/default/capture-agent-manager-gunicorn-upstart.conf.erb
```diff
@@ -0,0 +1,19 @@
...
+setgid <%= @user %>
+chdir /home/<%= @user %>/sites/<%= @name %>
```
lbjay: this should ensure that the root app dir is in `PYTHONPATH`

lbjay:
There's some work still to do re: logging. Currently the settings in cadash's `logging.yml` send regular log output to `stdout` which means it will end up in `/var/log/upstart/capture-agent-manager-gunicorn.log`. You also have an error log handler in there that i'm not sure is doing anything. I think we should update `logging.yml` to use a [SysLogHandler](https://docs.python.org/2/library/logging.handlers.html#sysloghandler) but we can work that out later.

lbjay:
Closing this PR as superseded by the new CA manage UI coming from @phoebem

PR #113, opened 2016-11-01
add `retries` to setting of cloudwatch retention policy

lbjay:
this avoids the race condition that can occur when AWS is still dealing
with the preceeding log group creation command


libraries/default.rb
```diff
@@ -486,6 +486,8 @@ def configure_cloudwatch_log(log_name, log_file, datetime_format)
 
       execute 'set log group retention policy' do
         command %Q|aws logs put-retention-policy --region #{region} --log-group-name #{log_group_name} --retention-in-days #{retention_days}|
+        retries 3
+        retry_delay 10
```
karendolan: Is that 10 seconds for the delay?

lbjay:
Yes.

PR #117, opened 2016-11-15
guard against executing cloudwatch log setup on local clusters

lbjay:


lbjay:
This branch/PR has been updated to address the issue reported in MATT-2238

PR #121, opened 2017-01-11
MATT-2232-L01-stream-id Configured L01 to point to the second Akamai id

rute-santos:
Configuration for L01 classroom to support the secondaries, which still have MHPearl 2.0, or in the case we rollback the primaries.


nmaekawa:
approved!

PR #124, opened 2017-02-10
MATT-2231 require login for annots summary

karendolan:
This goes with companion MATT-2231 pull in HUDCE MH bitbucket

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,6 @@ bin/*
 
 doc/
 .yardoc/
+
+.DS_Store
+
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -20,3 +20,6 @@ bin/*
 
 doc/
 .yardoc/
+
+.DS_Store
```
lbjay: fyi, for system-specific stuff you can create a global ~/.gitconfig file and then run

`git config --global core.excludesfile '~/.gitignore'`

PR #125, opened 2017-03-06
Add request time value to nginx access log events

lbjay:
Tested on a local vagrant cluster

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## TO BE RELEASED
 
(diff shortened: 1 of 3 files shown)


karendolan:
Hi Jay, can you also add in $upstream_response_time? In theory $response_time loosely equates to network throughput, but $upstream_response_time is if the server is being slow. 

Do you know if adding these things slows down nginx logging or the percentage of impact that logging these things causes nginx?

lbjay:
Ahh, good catch. I will do that (although it's not clear to me the values will differ significantly). Adding extra variables to the log output should have zero impact on nginx.

PR #126, opened 2017-03-13
adding request_time log format to engage proxy conf

lbjay:
this is a hotfix change that corresponds to an existing live edit


karendolan:
:thumbsup

PR #128, opened 2017-04-05
remove custom nodejs install recipe

lbjay:


karendolan:
@lbjay Does anything break if nodejs is removed from recipes/install-nodejs.rb?

lbjay:
I don't understand the question. You mean remove the recipe entirely?

karendolan:
yes

PR #130, opened 2017-05-08
Add GC logging and options for G1 garbage collection & YourKit jvm profiler agent

lbjay:
This PR add GC logging to our MH java start options. This is unobtrusive and will give us a bit more insight into JVM behavior.

The PR also adds a couple of cluster config options that allow to things.

* enabling the G1 garbage collection algorithm. Good overview of that [here](http://product.hubspot.com/blog/g1gc-fundamentals-lessons-from-taming-garbage-collection).
* including the [YourKit](https://www.yourkit.com/java/profiler/features/) java profiling agent to the MH startup options. This is intended only for enabling on a dev cluster and has no effect if not enabled.

### DIFF
--- a/templates/default/matterhorn-harness.erb
+++ b/templates/default/matterhorn-harness.erb
@@ -145,7 +145,19 @@ NEWRELIC="-javaagent:${NEW_RELIC_JAR} $NEWRELIC_CONFIG"
 NEWRELIC=""
 <% end %>
 
-MATTERHORN_OPTS="$NEWRELIC $GOSH $DEBUG $FELIX $GRAPHICS $JAVA $LOGGING $JMX $DISABLE_PHONEHOME $ADD_OPTS"
+<% if @enable_yourkit_agent %>
+YOURKIT="-agentpath:<%= @yourkit_dir %>/libyjpagent.so=dir=<%= @yourkit_dir %>,onlylocal"
+<% end %>
+
+GC="-Xloggc:${LOGDIR}/gcnormal.log -XX:+PrintGCDetails -XX:+PrintGCTimeStamps -XX:+PrintGCDateStamps"
+GC="${GC} -XX:+PrintGCApplicationStoppedTime -XX:+PrintGCApplicationConcurrentTime -XX:+PrintHeapAtGC"
+GC="${GC} -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=20 -XX:GCLogFileSize=32M"
+
+<% if @enable_G1GC %>
+GC="${GC} -XX:+UseG1GC"
+<% end %>
+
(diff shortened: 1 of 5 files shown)


karendolan:
@lbjay the JAVA="-Xms2048m -Xmx<%= @java_xmx_ram %>m -XX:MaxPermSize=1024m" are still passed. Does that interfere with G1 GC?

lbjay:
@karendolan no, the GC settings aren't dependent on any particular heap size params

templates/default/matterhorn-harness.erb
```diff
@@ -145,7 +145,19 @@ NEWRELIC="-javaagent:${NEW_RELIC_JAR} $NEWRELIC_CONFIG"
...
+
+GC="-Xloggc:${LOGDIR}/gcnormal.log -XX:+PrintGCDetails -XX:+PrintGCTimeStamps -XX:+PrintGCDateStamps"
+GC="${GC} -XX:+PrintGCApplicationStoppedTime -XX:+PrintGCApplicationConcurrentTime -XX:+PrintHeapAtGC"
+GC="${GC} -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=20 -XX:GCLogFileSize=32M"
+
+<% if @enable_G1GC %>
+GC="${GC} -XX:+UseG1GC"
+<% end %>
+
+MATTERHORN_OPTS="$NEWRELIC $GOSH $DEBUG $FELIX $GRAPHICS $JAVA $LOGGING $DISABLE_PHONEHOME $JMX $GC $YOURKIT $ADD_OPTS"
```
karendolan: @lbjay The "YOURKIT" var is part of the MATTERHORN_OPTS.  Does it need  the <% else %> YOURKIT="" <% end %> in case enable_yourkit_agent is false?

lbjay: it does not. bash won't complain if it's not defined. The `$NEWRELIC = ""` line above isn't really necessary either.

karendolan: Looks good to me.

templates/default/matterhorn-harness.erb
```diff
@@ -145,7 +145,19 @@ NEWRELIC="-javaagent:${NEW_RELIC_JAR} $NEWRELIC_CONFIG"
 NEWRELIC=""
 <% end %>
 
-MATTERHORN_OPTS="$NEWRELIC $GOSH $DEBUG $FELIX $GRAPHICS $JAVA $LOGGING $JMX $DISABLE_PHONEHOME $ADD_OPTS"
+<% if @enable_yourkit_agent %>
+YOURKIT="-agentpath:<%= @yourkit_dir %>/libyjpagent.so=dir=<%= @yourkit_dir %>,onlylocal"
```
lbjay: fyi, the `dir=..l` arg in that agentpath string tells yourkit where memory snapshots should be created.

PR #131, opened 2017-05-09
capture mh start-on-boot cron output

lbjay:
This is a very minor, no rush change. While looking into producer notification delays I noticed errors in the worker mail logs. Turns they're trying to email the output of the `/etc/init.d/matterhorn start` that runs on boot.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## TO BE RELEASED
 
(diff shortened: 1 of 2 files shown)


karendolan:
Great catch. Would it be better to append to syslog "... 1>>/var/log/syslog 2>&1"?

lbjay:
Actually better to do it the more standard way by piping to `/usr/bin/logger` which I forgot about.

PR #132, opened 2017-05-09
OPC- start and stop change to more sleep checks before retry start

karendolan:
This speeds up the start check back to 20 seconds, and prevents the daemon from restarting a starting OC prematurely.
Instead of restarting every 20 seconds, it checks every 20 seconds to see if OC is started. It sleeps 10 times (200 seconds) before trying a restart.


karendolan:
On the other hand, it might be easier to abandon the fuzzy check and use the official r/3.x https://bitbucket.org/opencast-community/matterhorn/src/9d4bcd7cbac9ff1ddd52de0967dfbd186b73dc7a/docs/scripts/service/etc-init.d-opencast?at=r%2F3.x&fileviewer=file-view-default

PR #133, opened 2017-05-10
Changes to jvm heap size settings

lbjay:
* bump max heap size `-Xxmx` for engage
* set initial heap size `-Xxms` for nodes based on ratio to `-Xxmx`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,8 @@
     "enable_yourkit_agent": true,
     ...
 ```
-    
+* set JVM `-Xxms` value based on ratio to`-Xxmx` depending on node type
+
 # v1.22.0 - 04/21/2017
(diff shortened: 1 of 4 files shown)


lbjay:
Here's a rundown of the changes and how it would impact prod:

* admin
  * max heap size: unchanged, 80% of total RAM == 47G on prod
  * initial heap size: was 2048m, will now be 10% of max == 4.7G on prod
* engage
  * max heap size: was 25% of total ram (14G), will now be 80% == 47G on prod
  * initial heap size: was 2048m, will now be 50% of max == 23.5G on prod
* workers
  * max heap size: unchanged, 25% of total RAM == 14G on prod
  * initial heap size: was 2048m, will now be 10% of max == 4.7G on prod

PR #136, opened 2017-06-19
MATT-2355-workspace-cleaner 

rute-santos:
Disable workspace cleaner on nodes depending on configuration. The idea is that only admin will run the ws cleaner in the future but, for this week, we need to disable it on all nodes.


karendolan:
:thumbsup:

PR #144, opened 2017-08-10
MATT-2388 open oauth user access to social reporting

karendolan:
This pull is required for the LTI social note tool


rute-santos:
Ok with me!

PR #146, opened 2017-08-21
recipes to pause/resume moscaler

lbjay:
Separates out the part of `install-moscaler` that creates the `cron_d` entries and creates new `moscaler-pause` and `moscaler-resume` recipes that delete and create the cron jobs respectively. The `moscaler-resume` recipe gets included at the end of `install-moscaler`. The pause/resume recipes will then be executed individually by new corresponding rake tasks in [mh-opsworks](harvard-dce/mh-opsworks).


karendolan:
looks good :thumbsup:

PR #149, opened 2017-09-08
MI-74: nginx config changes. includes:

lbjay:
* base nginx.conf from template, with `worker_processes` set
* relocate `request_type` log format definition to `nginx.conf`
* set `client_body_buffer_size` for admin nginx to prevent temp
  files being written for upload chunks

### DIFF
--- a/libraries/default.rb
+++ b/libraries/default.rb
@@ -529,6 +529,15 @@ module MhOpsworksRecipes
       configure_cloudwatch_log("nginx-access", "/var/log/nginx/access.log", "%d/%b/%Y:%H:%M:%S %z")
       configure_cloudwatch_log("nginx-error", "/var/log/nginx/error.log", "%d/%b/%Y:%H:%M:%S %z")
     end
+
+    def get_nginx_worker_procs
+      number_of_cpus = execute_command(%Q(nproc)).chomp.to_i
+      if admin_node? || engage_node?
+        [(number_of_cpus / 2), 4].max
+      else
+        4
+      end
+    end
   end
 
   module DeployHelpers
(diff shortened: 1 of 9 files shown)


libraries/default.rb
```diff
@@ -529,6 +529,15 @@ def configure_nginx_cloudwatch_logs
       configure_cloudwatch_log("nginx-access", "/var/log/nginx/access.log", "%d/%b/%Y:%H:%M:%S %z")
       configure_cloudwatch_log("nginx-error", "/var/log/nginx/error.log", "%d/%b/%Y:%H:%M:%S %z")
     end
+
+    def get_nginx_worker_procs
+      number_of_cpus = execute_command(%Q(nproc)).chomp.to_i
+      if admin_node? || engage_node?
+        [(number_of_cpus / 2), 4].max
```
lbjay: prod engage and admin have 32 cpus, so this would be 16. dev clusters would get a minimum of 4 (which is the current default setting)

templates/default/nginx-proxy.conf.erb
```diff
@@ -20,6 +16,8 @@ server {
   proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
 
   client_max_body_size 102400m;
+  # upload service file chunks are 100mb
+  client_body_buffer_size 101m;
```
lbjay: upload chunks should be exactly 100mb but an extra 1mb can't hurt, right?

templates/default/nginx.conf.erb
```diff
@@ -0,0 +1,34 @@
+user www-data;
```
lbjay: for this i took the default config that ships with nginx v1.4.6 and removed all the commented out sections.

rute-santos:
Looks good to me!

PR #151, opened 2017-09-14
MI-73: Capture Agent logs to Cloudwatch.

lbjay:
This adds a new recipe to be run on the Utility node that sets up a script for rsyncing capture agent logs from each CA device represented in the CA status json output. The script also generates configuration files for the logs to be monitored and pushed by the Cloudwatch Log Agent daemon.

### DIFF
--- a/files/default/rsync_ca_cwlogs.sh
+++ b/files/default/rsync_ca_cwlogs.sh
@@ -1,0 +1,63 @@
+#!/usr/bin/env bash
+
+set -e
+
+# get the list of ca hosts
+ca_list=`curl -s -u $CA_STATUS_USER_PASS $CA_STATUS_URL | jq -r '.[].address'`
+ca_count=`echo $ca_list | wc -w`
+echo "Fetched $ca_count entries"
+
+# rsync the logs
+for ca_host in $ca_list; do
+    echo "fetching from $ca_host"
+    rsync -e "ssh -i $CA_KEY_FILE" -azvh --quiet --inplace \
+        --exclude "messages.*" --exclude "mhpearl.log.*" \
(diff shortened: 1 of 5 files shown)


files/default/rsync_ca_cwlogs.sh
```diff
@@ -0,0 +1,63 @@
...
+
+# get the list of ca hosts
+ca_list=`curl -s -u $CA_STATUS_USER_PASS $CA_STATUS_URL | jq -r '.[].address'`
+ca_count=`echo $ca_list | wc -w`
+echo "Fetched $ca_count entries"
+
+# rsync the logs
+for ca_host in $ca_list; do
+    echo "fetching from $ca_host"
+    rsync -e "ssh -i $CA_KEY_FILE" -azvh --quiet --inplace \
```
lbjay: This `--inplace` flag is important as it tells rsync to update the existing, local file vs replace with the new one. Without it the behavior of the cw log agent is unpredictable.

files/default/rsync_ca_cwlogs.sh
```diff
@@ -0,0 +1,63 @@
...
+        --exclude "messages.*" --exclude "mhpearl.log.*" \
```
phoebem: How do you rotate the logs? Pearl logs are not rotated at midnight.(I can change it)

lbjay: no rotation needed on the local side. the rsync is only fetching `messages` and `mhpearl.log`. when a rotation happens on the device side the rsync will inplace update the local files with the new remote ones. there's no appending happening so the local logs don't grow and thus don't need to be rotated.

PR #152, opened 2017-10-05
MATT-2464 nginx error loglevel info for 400 bad request details

karendolan:
This changes log level from default error to info to get more details about the fileupload errors.
Ref http://nginx.org/en/docs/ngx_core_module.html#error_log

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## TO BE RELEASED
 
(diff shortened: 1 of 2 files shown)


karendolan:
Created a v1.28.0-rc1 on this commit for stage for verification (treating this like a hot fix)

lbjay (approved):

PR #156, opened 2017-10-25
install maxmind geolite2 db for zoom ip lookups

lbjay:
This is an update to earlier changes already merged into develop. Mona added geoip lookups for the zoom session data. Recipes updated to install the GeoLite2 database. Also modified the harvester install to use a virtualenv due to some issues with pip installing the python dependencies.

### DIFF
--- a/recipes/install-geolite2-db.rb
+++ b/recipes/install-geolite2-db.rb
@@ -1,0 +1,36 @@
+# Cookbook Name:: mh-opsworks-recipes
+# Recipe:: install-ua-harvester
+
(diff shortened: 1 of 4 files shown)


recipes/install-geolite2-db.rb
```diff
@@ -0,0 +1,36 @@
+# Cookbook Name:: mh-opsworks-recipes
+# Recipe:: install-ua-harvester
```
karendolan: Should that comment be #Recipe:: install-geolite2-db

lbjay: oof. yes. I also need to update `metadata.rb`. I'll fix in develop and make a new rc tag.

PR #158, opened 2017-12-01
disable `proxy_request_buffering` on admin nginx

lbjay:
This will allow direct streaming of uploads, bypassing the need for
temp buffers and or temp files

* install latest stable nginx from official nginx ubuntu package repo
* purges existing installation + config as config path structures have changed
* pins apt package to v1.12.*
* server configs now written to /etc/nginx/conf.d
* nginx service restart now handled the "right way"

### DIFF
--- a/CHANGELOG-5-x.md
+++ b/CHANGELOG-5-x.md
@@ -5,6 +5,22 @@ This log is structured differently from the base `CHANGELOG.md`.
 * There is no **TO BE RELEASED** section
 * Entries should be headed with just a date; no version numbers
 
+## 07/12/2018
+
+* *REQUIRES MANUAL RECIPE RUN*
+  install latest stable nginx & disable `proxy_request_buffering` for admin nginx to allow direct streaming of uploads. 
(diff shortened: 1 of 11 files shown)


lbjay:
I just tested a 52G upload and it went through in 23m. Obvs at some point down the road we'll want to do several concurrent uploads, but my selenium tools aren't written for the new admin.

karendolan:
@lbjay Is the reason to skip the nginx chunk buffering because the OC is manually handling chunk buffering and recovery? Nginx docs state the reason for its buffering is so nginx can switch to trying a  different upstream servers if it gets an error from the current one. Is that handing moot because the nginx is on the same node as admin, so if admin goes down so does that instance of nginx?

lbjay:
Maybe, sort of, and also just for performance reasons.

If by "OC manually handling chunk buffering" you mean the uploader javascript chunking the files into a series of upload requests I don't know how much of a factor that is. If that wasn't happening I think there'd be even more motivation to turn off request buffering. Imagine nginx having to receive and buffer an entire 40G file instead of just 100MB chunks.

The idea of switching to different upstream server is moot because we only have the one upstream.

PR #159, opened 2017-12-18
remove geoip and useragent parsing from logstash config 

lbjay:
This functionality will now be performed by the harvester, release v1.3.0, which will be released concurrently.


karendolan:
@lbjay this is pointed to old develop. Is this patch needed for the new oc-master or can it be closed?

lbjay:
Yeah, I'm going to close it. There was a corresponding change needed to the analytics pipeline that turned out to be more work than I thought.

PR #160, opened 2018-01-26
OPC MATT-2406 nodes that do not dispatch should not perform hearbeats

karendolan:
To reduce unnecessary internode ping flurry.

### DIFF
--- a/libraries/default.rb
+++ b/libraries/default.rb
@@ -700,7 +700,7 @@ module MhOpsworksRecipes
     end
 
     # nodes that don't dispatch should also not do hearbeat service health checks
-    def set_service_registry_dispatch_interval(current_deploy_root)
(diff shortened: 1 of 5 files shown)


libraries/default.rb
```diff
@@ -670,13 +670,15 @@ def install_otherpubs_service_series_impl_config(current_deploy_root)
       end
     end
 
+    # nodes that don't dispatch should also not do hearbeat service health checks
```
rute-santos: Change the method name? e.g. configure_service_registry?

karendolan: @rute-santos  I'm not sure what you mean by change the name.

rute-santos: Sorry, the comment is related to the line below :) The method is 'set_service_registry_dispatch_interval', but now we are configuring more than the dispatch interval.

PR #161, opened 2018-01-26
OPC MATT-2406 nodes that do not dispatch should not perform service check

karendolan:
Only the admin node needs to perform intra-node service checks since it also does the dispatching

### DIFF
--- a/libraries/default.rb
+++ b/libraries/default.rb
@@ -703,13 +703,15 @@ module MhOpsworksRecipes
       end
     end
 
+    # nodes that don't dispatch should also not do hearbeat service health checks
     def set_service_registry_dispatch_interval(current_deploy_root)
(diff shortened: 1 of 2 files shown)


karendolan:
@lbjay This pull has already been QA'd and I'd like to include it in the curret RC for stage. I'm going to run the CI tests a couple more times to verify that its not related to the code. Can you review the pull?

karendolan:
```
21.55s$ bundle install --jobs=3 --retry=3 --deployment --path=${BUNDLE_PATH:-vendor/bundle}
Fetching gem metadata from https://rubygems.org/.........
yard-chef-2.0.0 requires ruby version >= 2.2.0, which is incompatible with the
current version, ruby 2.0.0p648
```

PR #162, opened 2018-04-05
Oc opsworks 5.x recipes new

karendolan:
@rute-santos and @lbjay  I made this branch when we were testing develop-5.x-new. It' updates our custom config to better match OC develop. It doesn't do too much dramatic, but re-arranges some areas to match the OC version. I've been using this in kdolan-dev-5x.

### DIFF
--- a/templates/default/custom.properties.erb
+++ b/templates/default/custom.properties.erb
@@ -14,6 +14,10 @@
 # Related options (like listening addresses) are located in etc/org.ops4j.pax.web.cfg
 org.opencastproject.server.url=http://<%= @hostname %>
 
+# The environment specified below will show up in the email subject when sending a notification about transcription
+# errors. Useful to differentiate among environments (dev, test, etc) and may be used by other services in the future.
+org.opencastproject.environment.name=<%= @stack_name %>
+
 
 ######### USER/AUTHENTICATION #########
 
@@ -39,6 +43,17 @@ org.opencastproject.admin.email=dev-notify@lists.dce.harvard.edu
 org.opencastproject.security.digest.user=<%= @rest_auth[:user] %>
 org.opencastproject.security.digest.pass=<%= @rest_auth[:pass] %>
 
+# The number of times to retry a request if the nonce expires.
+#org.opencastproject.security.digest.nonce.retries=12
+
+# The configuration property specifying the minimum amount of time in seconds wait before retrying a request after
+# a nonce timeout. Default is 300 seconds (5 minutes).
+#org.opencastproject.security.digest.nonce.base.time=300
+
+# The maximum amount of time to wait in addition to the base time for a random generator to add after a nonce timeout
+# so that requests that timeout won't all try again at exactly the same time. Default is 300 seconds (5 minutes).
+#org.opencastproject.security.digest.nonce.variable.time=300
+
 
 ######### STORAGE #########
 
@@ -49,7 +64,8 @@ org.opencastproject.shared_storage.dir=<%= @shared_storage_root %>
 # #DCE This is the local node storage e.g. solr indices
 org.opencastproject.storage.dir=<%= @local_workspace_root %>
 
-# The path to the archive repository
+# The path to the asset manager directory
+# Default: ${org.opencastproject.storage.dir}/archive
 org.opencastproject.episode.rootdir=${org.opencastproject.shared_storage.dir}/archive
 
 # The path to the repository of files used during media processing.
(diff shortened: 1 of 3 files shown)


templates/default/custom.properties.erb
```diff
@@ -39,6 +43,17 @@ org.opencastproject.admin.email=dev-notify@lists.dce.harvard.edu
 org.opencastproject.security.digest.user=<%= @rest_auth[:user] %>
 org.opencastproject.security.digest.pass=<%= @rest_auth[:pass] %>
 
+# The number of times to retry a request if the nonce expires.
+#org.opencastproject.security.digest.nonce.retries=12
+
+# The configuration property specifying the minimum amount of time in seconds wait before retrying a request after
+# a nonce timeout. Default is 300 seconds (5 minutes).
```
lbjay: What is the nature of these requests? Just asking because 5m seems like a long time to wait before retrying.

templates/default/custom.properties.erb
```diff
@@ -61,9 +77,9 @@ org.opencastproject.workspace.rootdir=${org.opencastproject.shared_storage.dir}/
 # The location to store uploaded static files such as images and videos.
 org.opencastproject.staticfiles.rootdir=${org.opencastproject.shared_storage.dir}/staticfiles
 
-# The ID of the default workflow definition to run when media are ingested
-org.opencastproject.workflow.default.definition=ng-schedule-and-upload
-
+# Location of the temporary directory to build zip archives.
+# Default: ${org.opencastproject.storage.dir}/tmp/zip
+org.opencastproject.workflow.handler.workflow.ZipWorkflowOperationHandler.tmpdir=<%= @shared_storage_root %>/zip-tmp
```
lbjay: How big are these zip archives and how frequently are they created? Just trying to ascertain if doing zip operations over NFS is the best choice. Are they big enough that the (faster) local instance storage isn't safe or practical?

templates/default/custom.properties.erb
```diff
@@ -221,34 +221,36 @@ org.opencastproject.solr.dir=${org.opencastproject.storage.dir}/solr-indexes
...
+org.opencastproject.file.repo.url=<%= @admin_url %>
+
+# The max number of ingests to allow at the same time. If more ingests try than the max they will receive service
+# unavailable. A value of 0 means that the server will accept all ingests.
+# Default: 0
+#org.opencastproject.ingest.max.concurrent=0
+
+# The maximum number of concurrent files to ingest from the inbox directory
+# Default: 1
+#org.opencastproject.inbox.threads=1
```
lbjay: Do we use the inbox functionality enough that allowing multiple concurrent ingests helps anything?

templates/default/custom.properties.erb
```diff
@@ -301,10 +304,13 @@ org.opencastproject.staticfiles.webserver.enabled=false
 # for the uploaded static file. If not set Opencast uses endpoints to serve and secure the files.
 #org.opencastproject.staticfiles.webserver.url=${org.opencastproject.server.url}/staticfiles/
 
-# Timeout for capture agent status, in minutes.
-# Capture agents which have not sent status updates for this period will be marked as offline.
-# Default: 120 minutes (2 hours)
-#org.opencastproject.capture.admin.timeout=120
+# This is the maximum allowable size in bytes for a file to be uploaded. If the property is missing or set to 0 it is
+# disabled.
+# Default is 1000000000 which is 1GB.
```
lbjay: This seems to contradict the previous comment. If the property is missing is it disabled or is it set to the default of 1GB?

templates/default/custom.properties.erb
rute-santos: We are not generating zips anymore. I believe this is here to better match the one from upstream OC.

templates/default/custom.properties.erb
rute-santos: We are not currently using it (it was used a lot for repubs in the past).

templates/default/custom.properties.erb
karendolan: @lbjay  I don't know the nuances of this timeout. I suspect its the suggested delay on an unexpected digest auth communication issue. Here is the code where that param is used. https://github.com/opencast/opencast/blob/develop/modules/kernel/src/main/java/org/opencastproject/kernel/security/TrustedHttpClientImpl.java#L403

templates/default/custom.properties.erb
karendolan: @lbjay @rute-santos 's right. This is just moving the existing param in the file to match OC's ordering.

templates/default/custom.properties.erb
karendolan: @lbjay you're right. That's a typo in OC text. The default is 1GB if the param is missing. Only if set to 0 is the max size disabled.


  /** The maximum file size to allow to be uploaded in bytes, default 1GB */
private long maxUploadSize = 1000000000;

https://github.com/opencast/opencast/blob/develop/modules/static-file-service-api/src/main/java/org/opencastproject/staticfiles/endpoint/StaticFileRestService.java#L117

    if (cfgMaxUploadSize.isSome())
         maxUploadSize = Long.parseLong(cfgMaxUploadSize.get());

https://github.com/opencast/opencast/blob/develop/modules/static-file-service-api/src/main/java/org/opencastproject/staticfiles/endpoint/StaticFileRestService.java#L149

lbjay (approved):

PR #168, opened 2018-07-13
Disable tls1 sslv3 in nginx ssl protocols config

lbjay:


karendolan:
The pull looks reasonable to me. The QA for it will be a little painful :)

PR #171, opened 2018-09-05
t/MI-107: set rds write iops alert threshold based on allocated storage

lbjay:


lbjay:
Tested and confirmed on staging

PR #173, opened 2018-09-10
t/MI-106: improvements to activemq installation/logging

lbjay:
Changes:
* installs under the /opt/opencast volume (vs root volume) for persistence and increased storage capacity
* specify only TLSv1.1/2 protocols as TLSv1.0 is insecure/deprecated
* I left the log4j config mostly alone but increased the max file size to 10mb + keep 30 (up from 1mb + keep 5), a 60x increase in retention
* bump the -Xmx to 2048 for instances w/ > 16G of ram
* send activemq.log to cloudwatch


rute-santos:
Looks good to me.

PR #178, opened 2018-10-17
OPC-149 config to allow 5x combo pub list to pull from 1x

karendolan:
OPC-149 config to allow 5x combo pub list to pull from 1x engage server.

Includes documenting the params in otherpubs config template, adding an IGNORE-ME default value to alert otherpubs service that the params have not been intentionally set.

Note: there is an almost identical pull for the 1x to pull from 5x.

PR #179, opened 2018-10-17
OPC-149 config to allow 1x combo pub list to pull from 5x

karendolan:
OPC-149 config to allow 1x combo pub list to pull from 5x engage server.

Includes documenting the params in otherpubs config template, adding an IGNORE-ME default value to alert otherpubs service that the params have not been intentionally set.

Note: there is an almost identical pull for the 5x to pull from 1x. https://github.com/harvard-dce/mh-opsworks-recipes/pull/178

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## TO BE RELEASED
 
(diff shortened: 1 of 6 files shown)


rute-santos:
Looks OK to me.

PR #180, opened 2018-10-18
OPC-202: Fixed typo in assets path in mh_default_org.xml

rute-santos:
Fixed typo that was preventing producers to download media.


karendolan:
Looks great! 👍

PR #181, opened 2018-11-09
OPC-219: enable java debugging

lbjay:
This includes removing creation of obsolete `etc/opencast.conf` and allow enabling java debugging via `bin/setenv`


libraries/default.rb
```diff
@@ -864,27 +864,8 @@ def install_init_scripts(current_deploy_root, opencast_repo_root)
...
-
-      template %Q|#{current_deploy_root}/etc/opencast.conf| do
-        source 'opencast.conf.erb'
-        owner 'opencast'
-        group 'opencast'
-        variables({
-          opencast_root: opencast_repo_root + '/current',
-          node_profile: node_profile,
           opencast_log_directory: log_dir,
+          enable_newrelic: enable_newrelic_for_layer?(layer_name),
```
lbjay, line 868: this is not actually an added line; it shows up that way in the diff as a result of the `install_opencast_conf` function being removed.

templates/default/opencast-setenv.erb
```diff
@@ -42,6 +42,7 @@ NEWRELIC="-javaagent:${NEWRELIC_JAR} $NEWRELIC_CONFIG"
 NEWRELIC=""
 <% end %>
 
+KARAF_DEBUG=<% @enable_java_debug %>
```
lbjay: this will get set to an empty string if java debugging is not enabled, "1" or "true" otherwise, depending on how it was set in the cluster config

PR #183, opened 2018-11-16
OPC-226: redirect http -> https in engage nginx configuration

lbjay:
I thought this was going to be all or nothing, but turns out to be easy to do the https redirect only for certain paths. The problem with doing everything is that the internal cluster communications can't handle the 301 redirect.

`/admin-ng` is a given, but I'd like confirmation that anything under the `/engage/...` path is also OK to force https.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## TO BE RELEASED
 
+* OPC-226: redirect http -> https in engage nginx configuration
 * MI-125: rename maven cache archive to distinguish 5x vs 1x
 * OPC-219: remove creation of `etc/opencast.conf` and allow enabling java debugging via `bin/setenv`
 * MATT-2324: In 5x, it's Ok remove LTI process filter proxy workaround. Has companion OC 5x change.
--- a/templates/default/engage-nginx-proxy-conf.erb
+++ b/templates/default/engage-nginx-proxy-conf.erb
@@ -9,6 +9,12 @@ server {
   listen [::]:80 default_server ipv6only=on;
   directio 1M;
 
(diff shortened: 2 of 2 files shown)


karendolan:
Hi @lbjay, it's actually just the login page that requires the redirect to https. Links, like the links in Canvas, are already prefixed with "https". The engage path doesn't need to be encrypted because the session cookie is already encrypted.

This is the critical path:  http://opencast.dcex.harvard.edu/admin-ng/login.html

lbjay:
@karendolan but is there an argument against applying this to everything under `/admin-ng` and `/engage`? 

One thing it would do is prevent browsers from flagging things as "Not secure", e.g.
image

Also, based on grepping referrer values in the nginx logs, it looks like about 10% of player loads are still over http.

karendolan:
@lbjay As far as external browser clients go, I agree it would be good to force all to https. I do not know how this affects internal calls to those paths.

lbjay:
Right, that's what the change does: force https for only the "external" paths, `/admin-ng` and `/engage`. Internal calls between the nodes via http would not be affected. My first stab at this was a blanket redirect, but opencast's internal http requests didn't like that.

karendolan:
@lbjay Based on the proxy-5 tests, the 301 won't work for Canvas LTI integration.

lbjay:
OK good. So we should restrict it to only `/admin-ng/login.html`.  I'll update the PR.

karendolan:
This is already merged to "oc-master"
I think we are abandoning "oc-opsworks-5.x-recipes"?

lbjay:
Outdated pull. Will create a new branch/pr for the VPN whitelisting work that is also part of OPC-226.

PR #185, opened 2018-12-06
T/opc 201 live monitor recipe

karendolan:
@phoebem  I made this pull to be paired with your main AdminUI-NG Live monitor  https://bitbucket.org/hudcede/matterhorn-dce-fork/pull-requests/428/opc-201-live-monitor/ and also because I have a quick question on it...

### DIFF
--- a/templates/default/org.opencastproject.organization-mh_default_org.cfg.erb
+++ b/templates/default/org.opencastproject.organization-mh_default_org.cfg.erb
@@ -476,3 +476,7 @@ prop.admin.shortcut.general.help=?
 # Default: 5
 #
 #prop.admin.notification.duration.warning=5
+
+# #DCE OPC-201-live-monitor
+prop.dce.live.monitor.ca.stream=capture.device.live.resolution.960x270
+prop.dce.live.monitor.ca.stream.template=rtmp://cp398121.live.edgefcs.net/live/#{caName}-presenter-delivery.stream-960x270_1_200@355694


templates/default/org.opencastproject.organization-mh_default_org.cfg.erb
```diff
@@ -476,3 +476,7 @@ prop.admin.shortcut.general.help=?
 # Default: 5
 #
 #prop.admin.notification.duration.warning=5
+
+# #DCE OPC-201-live-monitor
+prop.dce.live.monitor.ca.stream=capture.device.live.resolution.960x270
+prop.dce.live.monitor.ca.stream.template=rtmp://cp398121.live.edgefcs.net/live/#{caName}-presenter-delivery.stream-960x270_1_200@355694
```
karendolan, line 482: @phoebem it that path good for dev, test, prod? It's the same path for all?

phoebem:
The first path is the one we actually need, as it is the key to the actual rtmp link in the structure returned from the capture agent.
I don’t want this to be baked in.

The second one is the default pattern which should never be exercised in production (we have 23 rooms now and 4 akamai ids), since it only
has a 1/4 chance of being correct if it were using a production id. 
It is actually using the dev akamai id, so if you have a fake CA (say lab-1) set up in the REST endpt with no CA attach, you can still see the 
video from the monitor.

karendolan:
@phoebem would it be better to make those 2 params into cluster configuration prams and popup them in that file via deploy like <%= @admin_hostname %>. Since the file is ready for that?

phoebem:
Those params are tied more to our capture agent configurations than to the clusters. The first one applies to all live streams, the second one is a backup from the way it was done before.
It doesn't care about admin host names, the caName applies to the capture agent name.

phoebem (approved):

PR #189, opened 2018-12-10
Changes related to mh-opsworks 5x

lbjay:
A few odds and ends related to the big mh-opsworks 5x changes PR


lbjay:
Deleting this as I've updated the branch name to reflect the JIRA ticket

PR #190, opened 2018-12-17
t/MI-137 5x changes

lbjay:
A few odds and ends related to the big mh-opsworks 5x changes PR


templates/default/nginx-proxy.conf.erb
```diff
@@ -19,7 +19,6 @@ server {
 
   # upload service file chunks are 100mb
   client_body_buffer_size 101m;
-  client_body_temp_path <%= @body_temp_path %>;
```
karendolan, line 22: @lbjay do you recall where that "100mb" upload service file chunk size is configured? Is it set by us within another recipe file? I ask because it seems fragile to remove the fallback when that buffer and the upload chunk size are so tightly coupled.

karendolan: On the other hand, if the system chokes becaues nginx tries to use temp strorage again, it's a heads up to recalibrate those sizes.

karendolan:
@lbjay this patch is pointed to oc-opsworks-5.x-recipes. Is that branch being depreciated? Has these changes already been merged to oc-master?

lbjay:
These changes are what became the `oc-master` branch. The original intention was to merge into `oc-opsworks-5.x-recipes` but probably not necessary at this point.

PR #192, opened 2019-01-03
Oc v2.0.0 rc2 with OPC-226 and typo fix from original 1x to 5x conversion

karendolan:
@lbjay This is the same as the OPC-266 pull but it fixes a 1x to 5x text-substitution typo in the proxy_redirect area of the Engage nginx. The fix resolves the issue where logging into the https://<DNS engage node>/... page changes back to http://<DNS engage node>... after login.
Tested with publish and retract on rute prod 5x migration cluster and social notes and pub list on harvard canvas sandbox (last two nav entries)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## TO BE RELEASED
 
 * MI-97: update mysql alarms for rds aurora cluster
+* OPC-226: redirect http -> https in engage nginx configuration
 * MI-125: rename maven cache archive to distinguish 5x vs 1x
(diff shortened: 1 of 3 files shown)


rute-santos:
Seems fine to me. If it has been tested, I think it's ok to merge. We need it next week :)

karendolan:
@rute-santos @lbjay , I just realized that public_engage_hostname is not enough for prod 5x which has both the old DNS and the new one. I'm updating to add the static old matterhorn.dce DNS back in along side the  config'd public_engage_hostname.

PR #195, opened 2019-01-11
autofs restart guard using wrong path value in zadara/external context

lbjay:


recipes/nfs-client.rb
```diff
@@ -47,7 +47,7 @@
 # Only restart if we don't have an active mount
 service 'autofs' do
   action :restart
-  not_if %Q|grep '#{storage_hostname}:#{export_root} ' /proc/mounts|
+  not_if %Q|grep '#{storage_hostname}:#{nfs_server_export_root} ' /proc/mounts|
```
lbjay, line 50: This is only an issue in zadara-backed clusters because in a dev cluster using a `storage1` node the `export_root` and `nfs_server_export_root` values are the same, i.e. `/var/opencast`

rute-santos:
Fine with me.

PR #200, opened 2019-02-08
MI-153: Attempting to address the 499s

lbjay:
* turn `proxy_buffering` back on and increase proxy buffers
* turn `proxy_ignore_client_abort` on
* unset extended read/send timeouts on engage
* enable nginx's status endpoint
* let nginx decide the right number of nginx workers

### DIFF
--- a/files/default/nginx-status.conf
+++ b/files/default/nginx-status.conf
@@ -1,0 +1,12 @@
+server {
+  listen 127.0.0.1:81;
+  server_name localhost;
+
+  access_log off;
+  allow 127.0.0.1;
+  deny all;
+
+  location /nginx_status {
+    stub_status;
+  }
+}
--- a/templates/default/elk-nginx-proxy-conf.erb
+++ b/templates/default/elk-nginx-proxy-conf.erb
@@ -1,5 +1,9 @@
 # This file is installed automatically by chef. Do not edit or delete.
 
+# increase size of default proxy buffers so that proxied content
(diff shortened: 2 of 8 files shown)


files/default/nginx-status.conf
```diff
@@ -0,0 +1,12 @@
...
+  server_name localhost;
+
+  access_log off;
+  allow 127.0.0.1;
+  deny all;
+
+  location /nginx_status {
+    stub_status;
+  }
+}
```
lbjay, line 12: This creates an endpoint that outputs some simple nginx stats. I lifted it from some DataDog setup instructions, but it's not datadog specific. I'd like to have it enabled in case we want to collect these metrics at some point.

Output example:
```
Active connections: 43 
server accepts handled requests
 7368 7368 10993 
Reading: 0 Writing: 5 Waiting: 38
```

rute-santos:
Looks good to me. I think we should include some uploads of big files in the QA tests.

templates/default/elk-nginx-proxy-conf.erb
```diff
@@ -1,5 +1,9 @@
 # This file is installed automatically by chef. Do not edit or delete.
 
+# increase size of default proxy buffers so that proxied content
+# doesn't get written to temp files
+proxy_buffers 16 16k;
```
lbjay, line 5: Sets the number/size of the buffers used for reading the upstream response.

PR #204, opened 2019-02-27
MI-159: Set `/etc/hostname` value to correct hostname

lbjay:


karendolan:
In the comment, you mean /etc/mailname ?

PR #205, opened 2019-03-01
t/OPC-314-editor-shortcut-display

phoebem:
Disable default help shortcut to display cheatsheet. This is the companion PR to the one of the same name in hudcede/matterhorn-dce-fork

### DIFF
--- a/templates/default/org.opencastproject.organization-mh_default_org.cfg.erb
+++ b/templates/default/org.opencastproject.organization-mh_default_org.cfg.erb
@@ -348,6 +348,7 @@ prop.show_embed_links=true
 #   - down
 #   - ins
 #   - del
+#   - plus
 # - any other key
 #   - should be referenceable by name like a, /, $, *, or =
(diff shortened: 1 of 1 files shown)


karendolan:
Looks good to me!

karendolan:
This has passed QA and the companion OPC-314 bitbucket pull is merged.

karendolan:
@phoebem I'm going to go ahead and merge this in for the Opencast 5.0.0/1.18.0-rc1 since the server side admin-ui is already expecting these params since 5.0.0/1.15.0.

karendolan:
FYI -  I removed the test to put an Engage service in a maintenance state while the other Engage services are active, because the endpoint only appears to allow putting the whole service into maintenance mode. The UI also does not have actions to add maintenance to services only on the server itself. The test scenario only works if the search service is up but the distribute service is down. Therefore, the current test for this fix is to make sure metasynch still works as usual.

PR #208, opened 2019-04-23
MI-162: new recipe to enable ubuntu advantage esm

lbjay:
This recipe will execute the `ubuntu-advantage enable-esm` command with the user/password combo from the cluster config's `ubuntu_advantage_esm` block. 

These values have already been added to `base-secrets.json` in both accounts. 

Existing cluster configs will need to be updated manually. 

Because the recipe gets included by `install-oc-base-packages` it will get run during normal cluster startup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## TO BE RELEASED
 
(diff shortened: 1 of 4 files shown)


rute-santos:
Looks good to me.

PR #212, opened 2019-07-05
  MI-171: nginx config reload needs to watch for changes to both the …

lbjay:
…ssl key *and* cert


karendolan:
Does the "reload" only trigger once, regardless of how many subscriptions it makes?

lbjay:
@karendolan yes, they queue up and get deduplicated during the chef run at the end.

karendolan:
👍

PR #216, opened 2019-08-05
fix new log4j2 syslog appender definition

lbjay:


libraries/default.rb
```diff
@@ -736,14 +736,16 @@ def install_opencast_log_configuration(current_deploy_root)
       oc_log_dir = node.fetch(:opencast_log_directory, '/var/log/opencast')
       oc_log_level = node.fetch(:opencast_log_level, 'INFO')
       dce_log_level = node.fetch(:dce_log_level, 'DEBUG')
+      hostname = node[:opsworks][:instance][:hostname]
       template %Q|#{current_deploy_root}/etc/org.ops4j.pax.logging.cfg| do
```
karendolan, line 740: Where does the "node[:opsworks][:instance][:hostname]" value come from? Is that something different than the cluster config?

karendolan:
The log4j2 ref docs showed capital "LOCAL3" vs lower case. But, it doesn't matter if it works Ok.

libraries/default.rb
lbjay: `node` is an internal chef data structure. it's available to all recipes.

lbjay:
Hmm, I think actually "LOCAL3" (uppercase) is the correct value. I see now that in loggly the messages for my jluker-5x cluster are showing up as "user-level messages" which means it thinks they're "LOCAL1" (i think), so maybe log4j2 isn't recognizing my lowercase "local3".

This facility thing is sort of irrelevant anyway. Here's an explanation of what they're for: https://success.trendmicro.com/solution/TP000086250-What-are-Syslog-Facilities-and-Levels

I think I originally chose "local3" with the idea that Opencast qualified as a "system daemon"? But you could probably argue they're user-level (LOCAL1). Anyway, like I said, it doesn't matter very much, but if we're going to specify it we should at least get the letter case right.

karendolan:
+1

PR #220, opened 2019-09-11
MI-126: adjust nfs availability cron and metric alarm

lbjay:
This adjusts the cron timing and alarm settings so that it's only triggered by two failures over two 60s periods to avoid false positives


karendolan:
👍

PR #224, opened 2020-01-23
OPC-446 Helix Google Sheets managed service config, and GoogleApi dat…

karendolan:
Required config for Helix googlesheet other pub service


rute-santos:
I have just one comment: please make parameter names consistent. I don't mind if they have underscores or camel case.

Examples:

In the cluster config:
sheet_id x defaultduration_min

In the service configuration:
edu.harvard.dce.otherpubs.googlesheet.dataStore (camel case) x edu.harvard.dce.otherpubs.googlesheet.header.zoomlink (all lower case) x 
edu.harvard.dce.otherpubs.googlesheet.defaultupdatedurationInMin (mixed).

There's no need to fix this now, but please fix in v2.

PR #236, opened 2020-11-17
MI-193: wipe pip cache directory used by the cloudwatch logs agent installer

lbjay:
The agent installer doesn't clean up the pip wheels it builds during the
install and they collect in this cache and are eating up disk space


rute-santos:
Looks good to me.

PR #238, opened 2021-01-04
OPC-580 Redirect http to https when host is *.harvard.edu

rute-santos:
This work is in preparation for the new auth because it's desirable to only allow the auth session cookie be sent with https requests only.
The idea is to only redirect http requests with a .harvard.edu name to https so that http can still be used for vagrant or a dev cluster that doesn't use auth. Note that the current auth already needs a .harvard.edu host name to work.
I've added a separate engage protocol config because engage host name is used in several places.
I've also updated all OC templates that refer to engage so that they use the correct protocol.

### DIFF
--- a/templates/default/engage-nginx-proxy-conf.erb
+++ b/templates/default/engage-nginx-proxy-conf.erb
@@ -13,6 +13,16 @@ proxy_buffers 16 16k;
 # Opencast throw jetty EofExceptions
 proxy_ignore_client_abort on;
 
+# All traffic that comes to http://SOMETHING.harvard.edu will be redirected to https.
+# The "harvard.edu" was used to exclude: the internal host name (used by Opencast job dispatching),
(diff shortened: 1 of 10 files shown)


templates/default/engage-nginx-proxy-conf.erb
```diff
@@ -13,6 +13,16 @@ proxy_buffers 16 16k;
 # Opencast throw jetty EofExceptions
 proxy_ignore_client_abort on;
 
+# All traffic that comes to http://SOMETHING.harvard.edu will be redirected to https.
+# The "harvard.edu" was used to exclude: the internal host name (used by Opencast job dispatching),
+# dev clusters that do not bother to set auth/ssl up, local vagrant.
```
lbjay, line 18: This comment could be more clear

lbjay (approved):
Code-wise LGTM.

karendolan (approved):
looks good!

PR #239, opened 2021-01-06
OPC-554: Changes for the new porta auth system

rute-santos:
Configuration for adapter and new auth systems.
Configuration for listener to series metadata update and service to push to porta.


rute-santos:
I will create a new PR for this.

PR #240, opened 2021-01-27
OPC-581: Video export

rute-santos:
Send email to producer with a list of links to presenter and presentation delivery files that can be downloaded for a limited amount of time.
Uses s3 presigned urls to the s3 bucket.


karendolan (approved):
Looks good

PR #245, opened 2021-06-23
OPC-554 Changes for the new porta auth system.

rute-santos:
Add configuration for porta auth service, auth migration adapter
Change search service logging temporarily to trace


karendolan (approved):
Looks good

PR #250, opened 2021-10-07
OPC-637 Make Metasynch distributiion configurable

karendolan:
Companion pull to https://bitbucket.org/hudcede/matterhorn-dce-fork/pull-requests/619/opc-637-metasynch-download-distribution-to

### DIFF
--- a/templates/default/custom.properties.erb
+++ b/templates/default/custom.properties.erb
@@ -117,6 +117,9 @@ org.opencastproject.download.url=<%= @engage_protocol %>://<%= @engage_hostname 
 # The directory to store media, metadata, and attachments for download from the engage tool
 org.opencastproject.download.directory=${org.opencastproject.shared_storage.dir}/downloads
 
+# DCE OPC-637 Make Metasynch distributiion configurable
+# Central access to download distribution service type to use (i.e. aws.s3 or download)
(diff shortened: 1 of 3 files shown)


templates/default/custom.properties.erb
```diff
@@ -117,6 +117,9 @@ org.opencastproject.download.url=<%= @engage_protocol %>://<%= @engage_hostname
 # The directory to store media, metadata, and attachments for download from the engage tool
 org.opencastproject.download.directory=${org.opencastproject.shared_storage.dir}/downloads
 
+# DCE OPC-637 Make Metasynch distributiion configurable
+# Central access to download distribution service type to use (i.e. aws.s3 or download)
+org.opencastproject.distribution.download.type=<%= @distribution_type %>
```
rute-santos, line 122: I like the idea of this being a "global" configuration that more than one service can use. I was going to ask you to change the LiveSchedulerService to use this, but then I remembered that the code is for upstream too so maybe at a later time when we can push this upstream?

rute-santos (approved):

templates/default/custom.properties.erb
karendolan: A later time!

PR #251, opened 2021-10-13
Transition to aws linux

lbjay:
Leaving these mostly as separate commits so it's easier to look at the individual changes, but I will squash prior to merge.


Berksfile
```diff
@@ -8,4 +8,6 @@ cookbook 'activemq', '~> 2.0.4'
 cookbook 'java', '1.47'
 cookbook 'maven', '~> 2.2.0'
 
+cookbook 'opsworks_initial_setup', path: 'opsworks_initial_setup/'
```
lbjay, line 20: Our recipe repo now includes this 2nd "cookbook". It extends a built-in Opsworks cookbook and overrides the `package_ntpd` recipe. This was the only workaround I could figure out for a buggish thing with opsworks and amazon linux. The instructions in [Set the time for your Linux instance](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/set-time.html) say to uninstall ntpd and install cronyd instead. Well, the built-in opsworks cookbook insists on installing ntpd during instance setup (which conflicts with cronyd. So we need to override that ntpd recipe and make it a no-op.

Berksfile.lock
```diff
@@ -9,6 +9,8 @@ DEPENDENCIES
   oc-opsworks-recipes
     path: .
     metadata: true
+  opsworks_initial_setup
+    path: opsworks_initial_setup
```
lbjay, line 13: including our cookbook extension for the berks packaging

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## TO BE RELEASED
 
+* MI-197: Elasticsearch package repo keyserver has changed
```
lbjay: These additions are still from an earlier PR. I need to update this.

attributes/default.rb
```diff
@@ -1,12 +1,8 @@
 
-# Java
-default['java']['install_flavor'] = 'openjdk'
-default['java']['jdk_version'] = '8'
-
 # Maven
 default['maven']['version'] = '3.3.9'
 default['maven']['setup_bin'] = true
-default['maven']['mavenrc']['opts'] = '-Dmaven.repo.local=/root/.m2/repository -Xms1024m -Xmx1024m -XX:PermSize=256m -XX:MaxPermSize=256m'
+default['maven']['mavenrc']['opts'] = '-Dmaven.repo.local=/root/.m2/repository -Xms1024m -Xmx1024m'
```
lbjay, line 5: removing some deprecated options

files/default/bash.bashrc
```diff
@@ -1,98 +1,94 @@
-# System-wide .bashrc file for interactive bash(1) shells.
+# /etc/bashrc
```
lbjay, line 1: Here i'm taking the default `.bashrc` from amazon linux and trying to incorporate all the custom stuff DanCP added to the ubuntu version

libraries/default.rb
```diff
@@ -19,20 +19,14 @@ def get_memory_limit
...
-      # We really want to be able to timeout and retry installs to get faster package
-      # mirrors. This is an annoying quirk with the ubuntu package mirror repos.
-      #
-      execute "install #{name}" do
-        environment 'DEBIAN_FRONTEND' => 'noninteractive'
-        command %Q|apt-get install -y #{name}|
-        retries 5
-        retry_delay 15
-        timeout 180
+    def install_package(names, opts="")
```
lbjay, line 29: we call this method with multiple package names in several recipes. Because of a quirk in `yum` we have to install each package with a separate command. If a package is already installed yum will exit with an non-zero status code (failure), so if your command is `yum install -y foo bar baz` and `foo` is already installed, yum will fail and `bar` and `baz` will never get installed.

libraries/default.rb
```diff
@@ -477,20 +462,12 @@ def git_repo_url(git_data)
...
           'es_data_path' => "/vol/elasticsearch_data",
           'es_enable_snapshots' => true,
-          'logstash_major_version' => '1:2.4',
-          'logstash_repo_uri' => 'http://packages.elasticsearch.org/logstash/2.4/debian',
           'logstash_tcp_port' => '5000',
           'logstash_stdout_output' => false,
-          'kibana_major_version' => '4.6',
-          'kibana_repo_uri' => 'https://packages.elastic.co/kibana/4.6/debian',
-          'curator_major_version' => '3.5',
-          'curator_repo_uri' => 'http://packages.elastic.co/curator/3/debian',
```
lbjay, line 493: i've moved all these version specifiers into the recipe itself. They don't really need to be configurable.

libraries/default.rb
```diff
@@ -540,41 +517,6 @@ def create_ssl_cert(ssl_info)
       end
     end
 
-    def get_capture_agent_manager_info
```
lbjay, line 543: All this capture agent manager stuff is old Naomi code that isn't used.

libraries/default.rb
```diff
@@ -593,61 +535,6 @@ def get_moscaler_info
         }.merge(node.fetch(:moscaler, {}))
     end
 
-    def configure_cloudwatch_log(log_name, log_file, datetime_format)
```
lbjay, line 596: Configuring cloudwatch logs at the Opsworks stack level now.

libraries/default.rb
```diff
@@ -661,6 +548,27 @@ def is_using_local_distribution?
...
+        # don't if virtualenv python is already correct
+        not_if %Q!test -d #{venv_path} && #{venv_path}/bin/python -V | grep -q "$(python3 -V)"!
+      end
+      if !requirements_path.nil?
+        execute "install requirements from #{requirements_path}" do
+          command "#{venv_path}/bin/python -m pip install -r #{requirements_path}"
+          user user
+        end
+      end
+    end
```
lbjay: standardizing how python virtualenvs get created

libraries/default.rb
```diff
@@ -661,6 +548,27 @@ def is_using_local_distribution?
...
+
+    def dont_start_opencast_automatically?
+      dont_start = node.fetch(:dont_start_opencast_automatically, "false")
+      return ['true', '1'].include? dont_start.downcase
+    end
```
lbjay: this fixes an old issue written up in DEPLOY-50

metadata.rb
```diff
@@ -264,22 +277,6 @@
 * logstash is configured
 '
 )
-recipe(
```
lbjay, line 267: this was ubuntu-specific

metadata.rb
```diff
@@ -647,19 +644,6 @@
 * Create a cloudwatch alarm hitched up to the opencast availability metric
 '
 )
-recipe(
```
lbjay, line 650: bash is already the default shell

metadata.rb
```diff
@@ -1059,6 +913,7 @@
 '
 )
 depends 'nfs', '~> 2.1.0'
+depends 'line', '~> 0.6.3'
```
lbjay, line 918: dependency issue where we need to pin the version of a dependent cookbook for compatibility reasons

recipes/configure-elk-nginx-proxy.rb
```diff
@@ -8,10 +8,7 @@
 es_host = node[:opsworks][:instance][:private_ip]
 
 include_recipe "oc-opsworks-recipes::install-nginx"
-install_package('apache2-utils')
-
-install_nginx_logrotate_customizations
-configure_nginx_cloudwatch_logs
+install_package('httpd-tools')
```
lbjay, line 11: same package, different name

recipes/install-run-one.rb
```diff
@@ -0,0 +1,25 @@
+# Cookbook Name:: oc-opsworks-recipes
```
lbjay, line 1: We use the `run-one` command for a lot of our cronjobs. `run-one` is ubuntu specific, but it's just a shell script so I was able to steal it and install it with this recipe.

recipes/install-utils.rb
```diff
@@ -2,11 +2,19 @@
...
+packages = %Q|htop nmap iotop pv nethogs sysstat dstat tree jq fio iperf3 git rsyslog-gnutls curl autofs|
 install_package(packages)
 
+# seems as good a place as any for this
+cookbook_file "bash.bashrc" do
+	path "/etc/bashrc"
+	owner "root"
+	group "root"
+	mode "644"
+end
```
lbjay, line 15: this used to happen in an ubuntu-specific recipe that also set bash as the default shell. moved it here instead for lack of a better place.

recipes/update-package-repo.rb
```diff
@@ -1,30 +1,19 @@
...
+  command 'yum-config-manager --enable epel'
   retries 5
   retry_delay 15
 end
 
-execute 'update package repository' do
-  environment 'DEBIAN_FRONTEND' => 'noninteractive'
-  command 'apt-get update'
+# not strictly necessary as the standard yum install commands will refresh
+# the package list when needed, but this might save a time when many packages
```
lbjay: "might save time"

recipes/update-python.rb
```diff
@@ -0,0 +1,16 @@
...
+packages = %Q|python38 python38-devel openssl-devel libffi-devel bzip2-devel gcc|
+install_package(packages)
+
+execute 'upgrade pip' do
+	command "/usr/bin/python3 -m pip install -U pip"
+end
+
+execute 'install virtualenv' do
+	command "/usr/bin/python3 -m pip install -U virtualenv"
+end
```
lbjay: make sure we're using the latest python3 available (in amazon linux) and the latest pip and virtualenv libraries

karendolan:
Hi Jay, it looks Ok to me as far as I can tell. Should we build some dev clusters with it and let QA verify OC works as usual? Has that already been tried?
FYI -  I updated the CHANGELOG from another pull, so that needs to be fixed in this pull.

karendolan (commented):
Hi @lbjay, this is a duplicate of the comment ... I can't see any issues, without redoing all the research you've done on this to figure out these updates. Has this been built with OC in a dev cluster for QA to check that OC works as usual with it?

The CHANGELOG needs to be resolved because recent comment conflicts.

libraries/default.rb
```diff
@@ -1042,7 +1007,7 @@ def install_publish_1x_service_config(current_deploy_root, enabled, engage_url)
         group 'opencast'
         variables({
           publish_1x_engage_url: engage_url,
-          publish_1x_enabled: enabled 
+          publish_1x_enabled: enabled
```
phoebem: At some point we should remove all the 1x stuff.  We have been on 5x a long time.

phoebem (approved):
Do you have a QA ticket on the new cluster? The changes seem fine and the packages are probably compatible.

lbjay:
@karendolan i've built several dev clusters as i iterated. OC seemed fine to me but of course i didn't exercise everything. Next will be a dedicated QA cluster.

@phoebem i'll go back through and look for 1x things that could be stripped out. No QA ticket yet. That's the next step.

PR #252, opened 2021-11-01
Immersive Classroom Chat CORS updates

karendolan:
This is to solve the issues of: 
1. only ONE if statement can set headers, because it removes all the previously set headers
2. It's safer to set headers outside of a If statement when ever possible
3. Once a header is set its hard to change, so it's better to use variables when possible to prevent duplicate headers

This pull has been tested in kdolan-oc-5x with immersive player CHAT request (with a companion update to the immersive-classroom code)

### DIFF
--- a/templates/default/engage-nginx-proxy-conf.erb
+++ b/templates/default/engage-nginx-proxy-conf.erb
@@ -122,7 +122,7 @@ server {
   set $cors_vary "";
   set $cors_expose_headers "";
 
-  if ($http_origin ~* (https?://.*\.harvard.edu(:[0-9]+)?)) {
+  if ($http_origin ~* (https?://.*\.dcex.harvard.edu(:[0-9]+)?)) {
         set $cors_origin $http_origin;
         set $cors_cred   true;
         set $cors_header $http_access_control_request_headers;
--- a/templates/default/engage-nginx-proxy-conf.erb
+++ b/templates/default/engage-nginx-proxy-conf.erb
@@ -107,9 +107,38 @@ server {
   client_max_body_size 102400m;
   gzip on;
 
-  add_header 'Access-Control-Allow-Origin' '*';
-  add_header 'Access-Control-Allow-Credentials' 'true';
-  add_header 'Access-Control-Allow-Methods' 'GET, OPTIONS';
+  # Avoid the if-is-evil-in-location nginx issue
+  # https://stackoverflow.com/questions/27955233/nginx-config-for-cors-add-header-directive-is-not-allowed
+  # Defaults for
+  # add_header 'Access-Control-Allow-Origin' '*';
+  # add_header 'Access-Control-Allow-Credentials' 'true';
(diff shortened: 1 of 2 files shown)


karendolan:
@rute-santos FYI -  this works great for me for the prefight OPTIONS and the GET, *except* for the first check, when it goes to Porta. I think it's because it's racing the player load right now. I'm going to add a delay on CHAT to verify that the first Porta check for Annots won't strip all the nginx headers if the player already has all the cookies set.

templates/default/engage-nginx-proxy-conf.erb
```diff
@@ -107,9 +107,38 @@ server {
...
+  set $cors_origin "*";
+  set $cors_cred   "true";
+  set $cors_method "GET, OPTIONS";
+  # No defaults for these
+  set $cors_header "";
+  set $cors_xframe_option "";
+  set $cors_vary "";
+  set $cors_expose_headers "";
+
+  if ($http_origin ~* (https?://.*\.harvard.edu(:[0-9]+)?)) {
```
rute-santos: dcex.harvard.edu?

templates/default/engage-nginx-proxy-conf.erb
```diff
@@ -107,9 +107,38 @@ server {
...
-  add_header 'Access-Control-Allow-Methods' 'GET, OPTIONS';
+  # Avoid the if-is-evil-in-location nginx issue
+  # https://stackoverflow.com/questions/27955233/nginx-config-for-cors-add-header-directive-is-not-allowed
+  # Defaults for
+  # add_header 'Access-Control-Allow-Origin' '*';
+  # add_header 'Access-Control-Allow-Credentials' 'true';
+  # add_header 'Access-Control-Allow-Methods' 'GET, OPTIONS';
+  set $cors_origin "*";
+  set $cors_cred   "true";
+  set $cors_method "GET, OPTIONS";
```
rute-santos, line 118: Isn't the chat client going to use POST?

karendolan: Actually, it uses PUT. 
The `GET, OPTIONS` defaults listed here are not part of the CORS headers used for [dcex]harvard.edu. 
Look further down, the cors_methods variable is changed to `set $cors_method $http_access_control_request_method;`. This means any method that comes in from [dcex]harvard.edu can get the CORS headers on them. They don't have to be listed. We want PUT and POST to both be available because POST is used for the alias name, if we decide to go with that.

templates/default/engage-nginx-proxy-conf.erb
karendolan: Yes, It could be more specific to dcex. I think all reactjs apps will be deployed to that in prod.

karendolan: done!

rute-santos (approved):

PR #254, opened 2022-01-11
OPC-621 CORS support for Immersive Classroom Video Chat OPC-646

karendolan:
OPC-621 support for dce_porta cookies on GET from *.harvard.edu domains
OPC-621 moved if origin to the location section of server as required by nginx
Immersive Classroom Chat CORS - only ONE if statement can set headers, and existing ones need to be reset. Also, set defaults variables to prevent duplicate headers
IC-CORS peer reivew, reduce scope to *.dcex.harvard.edu vs *.harvard.edu

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## TO BE RELEASED
 
+* IC cors support, requires nginx update on engage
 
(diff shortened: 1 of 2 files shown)


karendolan:
Merging OPC-621 updates to OC-IC integration stable branch

PR #257, opened 2022-01-25
DEPLOY-64: enable TLS for postfix smtpd

lbjay:
- include the mailx package which provides the `mail` command for testing
- configure correct paths to postfix tls session cache dbs
- create a dummy key/cert file
- clean up some ubuntu-specific stuff from the postfix config
- turn on additional postfix tls logging

### DIFF
--- a/recipes/enable-postfix-smarthost.rb
+++ b/recipes/enable-postfix-smarthost.rb
@@ -2,10 +2,11 @@
 # Recipe:: enable-postfix-smarthost
 
 ::Chef::Recipe.send(:include, MhOpsworksRecipes::RecipeHelpers)
-install_package('postfix')
 
 smtp_info = node.fetch(:smtp_auth, {})
 hostname = node[:opsworks][:instance][:hostname]
+postfix_cert_path = "/etc/ssl/certs/postfix.pem"
+mail_name = stack_shortname
 
 template %Q|/etc/postfix/main.cf| do
   source 'postfix-main.cf.erb'
@@ -13,8 +14,11 @@ template %Q|/etc/postfix/main.cf| do
   group 'root'
   mode '644'
   variables({
+    mail_name: mail_name,
     hostname: hostname,
-    relay_host: smtp_info[:relay_host]
+    origin: "#{hostname}.localdomain",
+    relay_host: smtp_info[:relay_host],
+    postfix_cert_path: postfix_cert_path
   })
 end
 
@@ -30,13 +34,6 @@ template %Q|/etc/postfix/sasl_passwd| do
   })
 end
 
-file '/etc/mailname' do
-  mode '600'
-  owner 'root'
-  group 'root'
-  content "#{hostname}.localdomain\n"
-end
-
 execute 'postmap the sasl_passwd file' do
   command %Q|postmap hash:/etc/postfix/sasl_passwd|
   retries 5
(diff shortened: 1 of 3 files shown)


recipes/enable-postfix-smarthost.rb
```diff
@@ -49,6 +53,19 @@
...
+  retries 5
+  retry_delay 5
+end
+
+file postfix_cert_path do
+  mode '400'
+  owner 'root'
+  group 'root'
+end
+
```
lbjay, line 61: On debian/ubuntu every installation includes a set of "snakeoil" cert/key files for this purpose. On amazonlinux we need to create our own. This allows the postfix smtpd component to enable TLS handling on incoming mail. I'm not 100% why we need that since we're only handling internally-generated messages, but we set it up that way originally so I guess we'll keep it working.

templates/default/postfix-main.cf.erb
```diff
@@ -1,12 +1,5 @@
 # See /usr/share/postfix/main.cf.dist for a commented, more complete version
 
-
-# Debian specific:  Specifying a file name will cause the first
-# line of that file to be used as the name.  The Debian default
-# is /etc/mailname.
-#myorigin = /etc/mailname
-
-smtpd_banner =  ESMTP  (Ubuntu)
```
lbjay, line 9: "The text that follows the 220 status code in the SMTP greeting banner."
The default is `$myhostname ESMTP $mail_name` which is all we need.

templates/default/postfix-main.cf.erb
```diff
@@ -18,20 +11,21 @@ append_dot_mydomain = no
 readme_directory = no
 
 # TLS parameters
-smtpd_tls_cert_file=/etc/ssl/certs/ssl-cert-snakeoil.pem
-smtpd_tls_key_file=/etc/ssl/private/ssl-cert-snakeoil.key
+smtpd_tls_cert_file=<%= @postfix_cert_path %>
```
lbjay, line 14: Pointing to the cert/key file created by the recipe. You only need `smtpd_tls_cert_file` if the key and cert are in the same file.

templates/default/postfix-main.cf.erb
```diff
@@ -18,20 +11,21 @@ append_dot_mydomain = no
...
 smtpd_use_tls=yes
-smtpd_tls_session_cache_database = btree:/smtpd_scache
-smtp_tls_session_cache_database = btree:/smtp_scache
+smtpd_tls_loglevel = 1
+smtpd_tls_session_cache_database = btree:${data_directory}/smtpd_scache
+smtp_tls_session_cache_database = btree:${data_directory}/smtp_scache
```
lbjay, line 18: The original config wasn't broken but it was creating some warnings in the maillog like
```
Jan 25 11:08:00 admin1 postfix/tlsmgr[3689]: warning: request to update table btree:/smtpd_scache in non-postfix directory /
Jan 25 11:08:00 admin1 postfix/tlsmgr[3689]: warning: redirecting the request to postfix-owned data_directory /var/lib/postfix
```

recipes/enable-postfix-smarthost.rb
```diff
@@ -30,13 +34,6 @@
   })
 end
 
-file '/etc/mailname' do
-  mode '600'
-  owner 'root'
-  group 'root'
-  content "#{hostname}.localdomain\n"
-end
-
```
lbjay, line 39: `/etc/mailname` is a debian/ubuntu thing. We'll configure the postfix mail_name directly in the postfix config.

PR #265, opened 2022-04-11
DEPLOY-76: allow use of prebuilt oc artifacts during deployment

lbjay:
- admin, engage and worker deploy recipes look for `oc_prebuilt_artifacts`
  block in the cluster config and, if enabled, fetch and extract
  tgz files from s3
- the opencast `revision` value in the cluster's app source
  configuration is used to build the s3 object url
- any `/` characters in the revision value are replaced with `-`
- the basic functionality of the `deploy_revision` chef resource is not
  modified; the opencast source is still pulled via git because we rely
  on the `before_symlink` callback to run and perform all of the
  installation, initialization and configuration tasks

### DIFF
--- a/libraries/default.rb
+++ b/libraries/default.rb
@@ -19,13 +19,13 @@ module MhOpsworksRecipes
       80
     end
 
-		# this method is called with multiple package names in several recipes.
-		# Because of a quirk in yum we have to install each package with
-		# a separate command. If instead we attempted to install multiple with
-		# a single command, and one of the packagex was already installed,
-		# yum will exit with an non-zero status code (failure), so if your
-		# command is `yum install -y foo bar baz` and foo was already installed,
-		# yum would fail and bar and baz would never get installed.
+    # this method is called with multiple package names in several recipes.
+    # Because of a quirk in yum we have to install each package with
+    # a separate command. If instead we attempted to install multiple with
+    # a single command, and one of the packagex was already installed,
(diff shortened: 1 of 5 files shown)


libraries/default.rb
```diff
@@ -19,13 +19,13 @@ def get_memory_limit
...
-		# yum will exit with an non-zero status code (failure), so if your
-		# command is `yum install -y foo bar baz` and foo was already installed,
-		# yum would fail and bar and baz would never get installed.
+    # this method is called with multiple package names in several recipes.
+    # Because of a quirk in yum we have to install each package with
+    # a separate command. If instead we attempted to install multiple with
+    # a single command, and one of the packagex was already installed,
+    # yum will exit with an non-zero status code (failure), so if your
+    # command is `yum install -y foo bar baz` and foo was already installed,
+    # yum would fail and bar and baz would never get installed.
```
lbjay, line 28: this is just a formatting correction

libraries/default.rb
```diff
@@ -843,12 +843,12 @@ def install_init_scripts(current_deploy_root, opencast_repo_root)
...
-				owner 'root'
-				group 'root'
-				mode '644'
-			end
+      cookbook_file 'opencast_sysconfig' do
+        path '/etc/sysconfig/opencast'
+        owner 'root'
+        group 'root'
+        mode '644'
+      end
```
lbjay, line 851: formatting

PR #267, opened 2022-04-15
f/OPC-674-auto-trim config defaults+logging

phoebem:
Config file for autotrim - dark and quiet signal detection


lbjay:
seems fine to me. would be nice to have the new config file be consistent in using `" = "` vs `"="` in the setting assignment.

PR #277, opened 2022-08-02
Should use default in source

phoebem:
Config file doesn't require variables and shouldn't be in the template


phoebem:
This removes org.opencastproject.signaldetection.impl.SignalDetectionServiceImpl.cfg.erb from the deployment provisioning since it has no dependencies, and OC should use the default one in <OC branch>/etc instead

rute-santos (approved):

PR #280, opened 2022-08-29
OPC-773 Fix bug where http connections were never released when downloading from S3

rute-santos:
- Enable aws sdk metrics.
- Add warn logging for aws.
- Configure aws s3 temp collection cleanup.

### DIFF
--- a/templates/default/opencast-setenv.erb
+++ b/templates/default/opencast-setenv.erb
@@ -31,8 +31,11 @@
 # export KARAF_BASE        # Karaf base folder
 # export KARAF_ETC         # Karaf etc  folder
 # export KARAF_OPTS        # Additional Karaf options
-# export KARAF_DEBUG       # Enable debug mode
 # export KARAF_REDIRECT    # Enable/set the std/err redirection when using bin/start
+#
+# Debug options
+# export KARAF_DEBUG       # Enable debug mode
+# export JAVA_DEBUG_PORT   # Set debug port (defaults to 5005)
 
 KARAF_DEBUG=<%= @java_debug_enabled %>
 LOGDIR=<%= @opencast_log_directory %>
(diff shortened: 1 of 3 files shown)


templates/default/opencast-setenv.erb
```diff
@@ -31,8 +31,11 @@
 # export KARAF_BASE        # Karaf base folder
 # export KARAF_ETC         # Karaf etc  folder
 # export KARAF_OPTS        # Additional Karaf options
-# export KARAF_DEBUG       # Enable debug mode
 # export KARAF_REDIRECT    # Enable/set the std/err redirection when using bin/start
+#
+# Debug options
+# export KARAF_DEBUG       # Enable debug mode
+# export JAVA_DEBUG_PORT   # Set debug port (defaults to 5005)
```
karendolan, line 38: Is adding option to change the default debug port to allow the service stop & and start to do the jdwp check to see if the service is running or stopped?

templates/default/opencast-setenv.erb
```diff
@@ -43,6 +46,8 @@ GC=""
 #GC="${GC} -XX:+PrintGCApplicationStoppedTime -XX:+PrintGCApplicationConcurrentTime -XX:+PrintHeapAtGC"
 #GC="${GC} -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=20 -XX:GCLogFileSize=32M -XX:+UseG1GC"
 
-export EXTRA_JAVA_OPTS="$GC -Dorg.eclipse.jetty.server.Request.maxFormContentSize=1500000 -Dfile.encoding=UTF-8"
+# #DCE OPC-773 Publish AWS metrics
+export EXTRA_JAVA_OPTS="${EXTRA_JAVA_OPTS} -Dcom.amazonaws.sdk.enableDefaultMetrics $GC -Dorg.eclipse.jetty.server.Request.maxFormContentSize=1500000 -Dfile.encoding=UTF-8"
```
karendolan, line 50: Should we be publishing metrics from all clusters? Do you have docs on how to reference the metrics? Or is this to enable the logging in the opencast logs?

karendolan (approved):
Looks good to me, just a question about enabling aws metrics for all clusters

templates/default/opencast-setenv.erb
rute-santos: This came from upstream so I'm not sure.

templates/default/opencast-setenv.erb
rute-santos: This will publish the metrics to CloudWatch. The side effect is that they appear in the Opencast log if you set the log level to DEBUG (currently in WARN only).

PR #281, opened 2022-09-22
F/opc 737 thumbnail email

phoebem:
Added  production_management_email to template>default>org.opencastproject.organization-mh_default_org.cfg.erb for OPC-737 thumbnail


recipes/deploy-admin.rb
```diff
@@ -208,6 +210,7 @@
         stack_name: stack_name,
         workspace_cleanup_period: 86400,
         activemq_bind_host: activemq_bind_host,
+        production_management_email: production_management_email,
         distribution_type: distribution
```
karendolan, line 214: Does the param need to be added to custom.properties?

phoebem: This was already defined in custom.properties.erb, just not set by the scripts

# #DCE Production management email (OPC-668)
edu.harvard.dce.production.email=<%= @production_management_email %>

PR #259, opened 2022-02-14
OPC-647 Add immersive classroom url to custom.properties

rute-santos:
This is a companion pull to https://bitbucket.org/hudcede/matterhorn-dce-fork/pull-requests/644

PR #282, opened 2022-09-30
OPC-646 Immersive Classrooom integration stable 11

karendolan:
This is the IC-OC-Integration branch migrated to oc-master vs OC v11. 

It includes CORS support for  requests from *.harvard.edu addresses to enable the Porta Cookie to be sent in the Video Comments request from Immersive Classroom servers..  

The pull from https://github.com/harvard-dce/mh-opsworks-recipes/pull/259 was cherry-picked to f/OPC-646-IC-Integration-stable-11.

### DIFF
--- a/templates/default/custom.properties.erb
+++ b/templates/default/custom.properties.erb
@@ -388,3 +388,6 @@ edu.harvard.dce.production.email=<%= @production_management_email %>
 edu.harvard.dce.cas.login.url=https://www.pin1.harvard.edu/cas/login
 edu.harvard.dce.cas.service=<%= @cas_service %>
 
+# Immersive classroom base url. Used to push a notification when a video is replaced by its trimmed version.
+edu.harvard.dce.immersive.classroom.url=<%= @immersive_classroom_url %>
+
--- a/templates/default/engage-nginx-proxy-conf.erb
+++ b/templates/default/engage-nginx-proxy-conf.erb
@@ -107,9 +107,38 @@ server {
   client_max_body_size 102400m;
   gzip on;
 
-  add_header 'Access-Control-Allow-Origin' '*';
-  add_header 'Access-Control-Allow-Credentials' 'true';
-  add_header 'Access-Control-Allow-Methods' 'GET, OPTIONS';
+  # Avoid the if-is-evil-in-location nginx issue
+  # https://stackoverflow.com/questions/27955233/nginx-config-for-cors-add-header-directive-is-not-allowed
+  # Defaults for
+  # add_header 'Access-Control-Allow-Origin' '*';
+  # add_header 'Access-Control-Allow-Credentials' 'true';
+  # add_header 'Access-Control-Allow-Methods' 'GET, OPTIONS';
+  set $cors_origin "*";
+  set $cors_cred   "true";
+  set $cors_method "GET, OPTIONS";
+  # No defaults for these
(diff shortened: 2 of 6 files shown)


karendolan:
@rute-santos Even though it does no harm, should that IC reference from OC admin be removed from this pull? It is only needed for the OC re-trim call to IC feature which has not gone through QA.

templates/default/engage-nginx-proxy-conf.erb
```diff
@@ -107,9 +107,38 @@ server {
...
+  set $cors_origin "*";
+  set $cors_cred   "true";
+  set $cors_method "GET, OPTIONS";
+  # No defaults for these
+  set $cors_header "";
+  set $cors_xframe_option "";
+  set $cors_vary "";
+  set $cors_expose_headers "";
+
+  if ($http_origin ~* (https?://.*\.dcex.harvard.edu(:[0-9]+)?)) {
```
lbjay: did you look at alternatives to baking the hostname pattern into the code? We've managed to avoid this so far.

templates/default/custom.properties.erb
```diff
@@ -388,3 +388,6 @@ edu.harvard.dce.production.email=<%= @production_management_email %>
 edu.harvard.dce.cas.login.url=https://www.pin1.harvard.edu/cas/login
 edu.harvard.dce.cas.service=<%= @cas_service %>
 
+# Immersive classroom base url. Used to push a notification when a video is replaced by its trimmed version.
+edu.harvard.dce.immersive.classroom.url=<%= @immersive_classroom_url %>
```
lbjay, line 392: Will it cause problems for this to get the string 'None' when `immersive_classroom_url` isn't set?

rute-santos:
@karendolan I have no strong feelings. The ic url can stay as it causes no harm or be removed and added later.

templates/default/custom.properties.erb
karendolan: Hi Jay, good catch. This param is part of Rute's notify IC feature. Her OC feature code snippet below, shows that this is expecting a string or empty string. So, I'll change the default to this from 'None' to '' empty string.

```
+    // Skip if immersive classroom url was not set. This is OK for a dev cluster.
+    if (StringUtils.isEmpty(immersiveBaseUrl)) {
+      return createResult(Action.SKIP);
+    }
```

templates/default/engage-nginx-proxy-conf.erb
karendolan: Updated this to be generic. It now extracts the domain from the engage public host name. 
Also patched an issue with unescaped dots. this is what the rendered update looks like when deployed to dev Engage node.
`if ($http_origin ~* (https?://.*\.dcex\.harvard\.edu(:[0-9]+)?)) {`

templates/default/custom.properties.erb
karendolan: Changed to "" so the OC code will skip default param when not set in a cluster config.

karendolan:
@lbjay , I updated the pull from your 2 excellent comments.
@rute-santos , I left in the param for the notify IC feature, since it does no harm. I changed the default to "" to match what the OC code expects for default.

lbjay:
Looks great. Thanks for the reminder!

lbjay (approved):

PR #283, opened 2022-10-12
OPC-726: Remove legacy auth configuration (acid)

rute-santos:
Also includes adding the thread id to the logs.


karendolan (approved):
looks good

PR #285, opened 2022-11-04
OPC-696 HarvardKey/CAS configuration

rute-santos:
Note: I intend to remove the spring cas TRACE logging before going to production.

### DIFF
--- a/recipes/deploy-all-in-one.rb
+++ b/recipes/deploy-all-in-one.rb
@@ -165,7 +165,7 @@ deploy_revision "opencast" do
     install_opencast_log_management
     install_multitenancy_config(most_recent_deploy, public_hostname, public_hostname, public_engage_protocol)
     install_smtp_config(most_recent_deploy)
-    install_default_tenant_config(most_recent_deploy)
+    install_default_tenant_config(most_recent_deploy, false, public_hostname)
     install_porta_auth_service(
       most_recent_deploy, porta_auto_url, porta_auto_cookie_name, porta_auto_redirect_url, porta_auto_enabled
     )
--- a/templates/default/mh_default_org.xml.erb
+++ b/templates/default/mh_default_org.xml.erb
@@ -29,6 +29,7 @@
 
     <!-- Allow anonymous access to the login form -->
     <sec:intercept-url pattern="/" access="ROLE_ANONYMOUS" />
+    <sec:intercept-url pattern="/cas*" access="ROLE_ANONYMOUS" />
     <sec:intercept-url pattern="/admin-ng/login.html" access="ROLE_ANONYMOUS" />
     <sec:intercept-url pattern="/sysinfo/bundles/version" method="GET" access="ROLE_ANONYMOUS" />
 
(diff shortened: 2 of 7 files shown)


recipes/deploy-all-in-one.rb
```diff
@@ -165,7 +165,7 @@
     install_opencast_log_management
     install_multitenancy_config(most_recent_deploy, public_hostname, public_hostname, public_engage_protocol)
     install_smtp_config(most_recent_deploy)
-    install_default_tenant_config(most_recent_deploy)
+    install_default_tenant_config(most_recent_deploy, false, public_hostname)
```
karendolan, line 168: Maybe add a comment about forcing CAS to be disabled for DCE all-in-ones because all-in-one are usually used for quick or local development and do not keep a fixed long term public host name required for DCE HK CAS?

templates/default/mh_default_org.xml.erb
```diff
@@ -29,6 +29,7 @@
 
     <!-- Allow anonymous access to the login form -->
     <sec:intercept-url pattern="/" access="ROLE_ANONYMOUS" />
+    <sec:intercept-url pattern="/cas*" access="ROLE_ANONYMOUS" />
```
karendolan: Should this be /cas/* ?  I'm wondering if /cas* might include unintended potential future paths like /casting and /casual

karendolan: or maybe add both
````
 <sec:intercept-url pattern="/cas" access="ROLE_ANONYMOUS" />
 <sec:intercept-url pattern="/cas/**" access="ROLE_ANONYMOUS" />
```

templates/default/org.ops4j.pax.logging.cfg.erb
```diff
@@ -26,6 +26,10 @@ log4j2.logger.opencast.level = <%= @opencast_log_level %>
 log4j2.logger.hudce.name = edu.harvard.dce
 log4j2.logger.hudce.level = <%= @dce_log_level %>
 
+# TEMPORARY
+log4j2.logger.springcas.name = org.springframework.security
+log4j2.logger.springcas.level = TRACE
```
karendolan: would this be in prod, or only during stage QA?

karendolan (commented):
So LDAP would be disabled and CAS enabled in the config to support this? Does this support toggling between HK-CAS and LDAP, until they take down the ldap system?

templates/default/mh_default_org.xml.erb
rute-santos: I can try that. I will need to re-test.

recipes/deploy-all-in-one.rb
rute-santos: Is all-in-one even working? :)

templates/default/org.ops4j.pax.logging.cfg.erb
rute-santos: I re-read my comment above. It should be removed before going to PROD. I will do that next. Thanks for catching!

karendolan (approved):
Looks good

PR #286, opened 2022-12-05
OPC-831: Add stack name to organization properties

rute-santos:
So that the client can display the cluster name.

### DIFF
--- a/recipes/update-host-based-configurations.rb
+++ b/recipes/update-host-based-configurations.rb
@@ -7,6 +7,7 @@
 opencast_repo_root = node[:opencast_repo_root]
 
 production_deploy_root = opencast_repo_root + '/current'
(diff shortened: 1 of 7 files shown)


recipes/update-host-based-configurations.rb
```diff
@@ -7,6 +7,7 @@
 opencast_repo_root = node[:opencast_repo_root]
 
 production_deploy_root = opencast_repo_root + '/current'
+stack_name = stack_shortname
```
karendolan, line 10: Is stack_shortname already a param somewhere?

rute-santos: It's a function in libraries/defaults.rb.

karendolan (approved):
Looks good to me

PR #294, opened 2023-06-20
DEPLOY-86: update recipes to be compatible with RHEL7

lbjay:
Siginificant changes include:

- java 11 openjdk installed (instead of amazon corretto)
- installation and enabling of rhel7 "software collections" for python 3.8 and mysql 5.7
- switch to using systemd service units for activemq and opencast
- update chef service blocks to explicitly use systemd
- setting selinux to permissive, meaning it will log warnings but not block anything
- creating a common function for pip installing python packages
- fixed issue where karaf pid/lock directories weren't getting created prior to OC start
- crowdstrike recipe needed to use `yum localinstall` to prevent errors when package was already installed (eg as part of the AMI)
- disable ssl on the jdbc connection from OC to MySQL: The RDS instance is on a private subnet and all communication between OC and the RDS resources takes place within the VPC, so disabling SSL is OK
- fixing a bug in the ganglia monitoring service definition file causing hung opsworks configure lifecycle events


recipes/install-elasticsearch.rb
```diff
@@ -30,11 +30,13 @@
 end
 
 include_recipe "oc-opsworks-recipes::update-package-repo"
-install_package("java-1.8.0-openjdk java-1.8.0-openjdk-devel elasticsearch python-elasticsearch-curator")
+install_package("java-1.8.0-openjdk java-1.8.0-openjdk-devel elasticsearch")
+pip_install("elasticsearch-curator==3.5.1")
```
karendolan, line 34: elastic search requires java-1.8.0-openjdk not java-11-openjdk?

recipes/install-oc-base-packages.rb
```diff
@@ -5,36 +5,19 @@
...
+# Something in OC, possibly the spring framework stuff, requires
+# TLSv1.1 to be enabled for the jdbc connection handshake to work,
+# so that's all this file does: removes TLSv1.1 from the list
+# of disabled tls protocols
+cookbook_file 'java.security' do
+  path %Q|/usr/lib/jvm/java-11-openjdk/conf/security/java.security|
+  owner "root"
+  group "root"
+  mode "644"
 end
```
karendolan: There is a StackOverflow on a similar issue where one response was that they upgraded their mysql-connector-java to v5.1.49 to resolve the issue. We are on v5.1.43 and have been since 2017. This is probably because Opencast upstream removed mysql support in later versions and it's up to DCE to update that lib. 
Not sure if this will be the silver bullet for the TSL1.2 issue, but maybe worth trying. https://stackoverflow.com/questions/67332909/why-can-java-not-connect-to-mysql-5-7-after-the-latest-jdk-update-and-how-should/71631180#71631180

There is a Stackoverflow about TLS1.2 
3226732cb0e (Lars Kiesow         2017-08-30 21:59:49 +0200 1250)         <version>5.1.43</version>
```
      <!-- #DCE Use mysql -->
      <dependency>
        <groupId>mysql</groupId>
        <artifactId>mysql-connector-java</artifactId>
        <version>5.1.43</version>
      </dependency>
 ```

File: templates/default/custom.properties.erb [hunk 3]
```diff
@@ -127,7 +127,7 @@ org.opencastproject.db.jdbc.driver=com.mysql.jdbc.Driver
 #  .url=jdbc:h2:${org.opencastproject.storage.dir}/db
 #  .user=sa
 #  .pass=sa
-org.opencastproject.db.jdbc.url=jdbc:mysql://<%= @database[:host] %>/<%= @database[:database] %>
+org.opencastproject.db.jdbc.url=jdbc:mysql://<%= @database[:host] %>/<%= @database[:database] %>?autoReconnect=true&useSSL=false
```
karendolan, line 131: I'm curious why you added the useSSL=false here. This will impact the TLS version that the jdbc connection can support.

karendolan (commented):
I left some comments

recipes/install-oc-base-packages.rb
rute-santos: This rings a bell. I believe we tried that already?

rute-santos (approved):
Looks good to me.

recipes/install-elasticsearch.rb
lbjay: Yes, because the elasticsearch version in use is so old. :(

templates/default/custom.properties.erb
lbjay: I tried several suggestions I found from googling with no success and finally decided since the RDS cluster is running on a private subnet it was OK to just disable SSL. I think this happened **AFTER** i'd already introduced the alternative `java.security` config file, so I'm not 100% sure if the latter is still necessary. 

I'm going to go ahead and revisit this and try some additional things to make sure this is the best solution. For instance, that SO thread mentions some debugging options I didn't try. Also I guess in some cases the tls connection string parameter is `tlsVersion` instead of `enabledTLSProtocols`.

lbjay: I tried again with the same result. I was able to determine that with `useSSL=false` we no longer need to provide a modified `java.security` file. So that's saving a bit of complexity.

recipes/install-oc-base-packages.rb
lbjay: We did but I'm going to try one more time. It's possible there was another problem going on when I tried this previously.

lbjay: OK, I tried this again and it actually worked this time. I was able to add `enabledTLSProtocols=TLSv1.2` to the connection string instead of `useSSL=false`. HOWEVER, now I get a warning in the opencast startup output like:

```
WARN: Establishing SSL connection without server's identity verification is not recommended. According to MySQL 5.5.45+, 5.6.26+ and 5.7.6+ requirements SSL connection must be established by default if explicit option isn't set. For compliance with existing applications not using SSL the verifyServerCertificate property is set to 'false'. You need either to explicitly disable SSL by setting useSSL=false, or set useSSL=true and provide truststore for server certificate verification.```
```

It's just a warning but I don't know if I'm ok just ignoring it. For one thing it's noisy. I'm trying to read up on how to "provide truststore for server certificate verification" and it's all very confusing. I'm accustomed to all this stuff just working.

PR #299, opened 2023-09-13
remove unused and non-functional harvesting of zoom analytics

lbjay:


recipes/configure-ua-harvester.rb
```diff
@@ -13,14 +13,6 @@
 region = node[:opsworks][:instance][:region]
 es_host = node[:opsworks][:instance][:private_ip]
 max_start_end_span = elk_info.fetch('harvester_max_start_end_span', nil)
-zoom_key = elk_info['zoom_key']
-zoom_secret = elk_info['zoom_secret']
-loggly_info = node.fetch(:loggly, { token: '', url: '' })
-loggly_config = if loggly_info[:token] != ''
```
lbjay, line 19: loggly isn't used by the harvesting script. this is a remnant from long ago.

PR #301, opened 2023-10-23
service config should use the symlinked path to the `current` release

lbjay:
This is a laughably simple fix. I've tested it on rute-oc-5x-3 several times, going back and forth between new and previous releases. Also ensured that it doesn't cause issues when new instances are added and first brought online.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## TO BE RELEASED
 
+- `opencast.service` file should use path to the current deployment (vs hardcoded release path)
+
 # v5.0.3 - 10/04/2023
 
 - Clean pip cache just like we do for yum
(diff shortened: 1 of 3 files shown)


templates/default/opencast.service.erb
```diff
@@ -7,8 +7,8 @@ After=network.target
 After=remote-fs.target
 
 [Service]
-ExecStart=/<%= @opencast_root %>/bin/start-opencast server
-ExecStop=/<%= @opencast_root %>/bin/stop-opencast
+ExecStart<%= @opencast_root %>/bin/start-opencast server
```
rute-santos: Is the "=" supposed to be missing here?

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## TO BE RELEASED
 
+- `opencast.service` file should use path to the current deployment (vs hardcoded reelase path)
```
rute-santos: Typo reelase

templates/default/opencast.service.erb
lbjay: yikes, no. i was just trying to sneak in removing the extra `/`.

PR #302, opened 2023-11-29
OPC-940 Protect /search/lucene* endpoint.

rute-santos:


karendolan (approved):
looks good. I didn't know OC even had that endpoint

PR #303, opened 2024-01-04
OPC-855 Add redirect from old engage host name to new one, except for lti tools

rute-santos:


rute-santos:
Closing because this will not work with Opencast LTI.

PR #306, opened 2024-04-04
Protect iFrame embed to whitelist

karendolan:
NOTE: Leave this pull open in the case that non-Harvard site embed issues become a priority.
-----------------------------------------------------
Verified whitelist and harvard.edu cases in QA-1420.
I verified local dev localhost whitelist works running local IC.

### DIFF
--- a/libraries/default.rb
+++ b/libraries/default.rb
@@ -242,6 +242,11 @@ module MhOpsworksRecipes
       node.fetch(:vpn_ips, []) + ["127.0.0.1/32"]
     end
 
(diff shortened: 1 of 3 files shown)


karendolan:
@lbjay @rute-santos FYI  the issue with merging this is that we are afraid our videos are iFramed in non-harvard.edu sites and we didn't want to disturb that. It's a policy question whether to adopt this or not. This pull enables us to white-list non-harvard sites so they can embed, but we have to know about them to add then to the whitelist.

PR #308, opened 2024-05-30
Recipe changes for OC 15.x

rute-santos:


lbjay:
Looks like `mh_default_org.xml.erb` is still trying to make use of ldap configuration variables. it's in a section that's disabled if ldap is enabled, but still weird.

lbjay (approved):

PR #309, opened 2024-06-14
queued job metric script now gets value directly from endpoint via curl

lbjay:
also removed scripts and cron entries for unneeded worker metrics


rute-santos (approved):
Looks good.

PR #311, opened 2024-08-08
OPC-1017 Add Gather api url to config.

rute-santos:
Add Gather event api url to custom.properties.


karendolan (approved):
Looks good to me!

